PR #25, opened 2020-02-26
Test installation across different platforms.

95-martin-orion:
These tests install qsimcirq directly, without using docker. This verifies the most convenient way to use qsim, and avoids the [licensing issues](https://github.community/t5/GitHub-Actions/Why-is-Docker-not-installed-on-macOS/td-p/39364) with running Docker on macOS Github VMs.


95-martin-orion:
Based on offline discussion, we'll look into different ways of testing qsim on other platforms.

PR #43, opened 2020-03-11
Add objects for full state output.

### LINKED ISSUE #40: QSimSimulator: Can't access final state or set initial state
https://github.com/quantumlib/qsim/blob/24a9af400c3d9e4aac011cb8e5dc6b9e1ac4233b/qsimcirq/qsim_simulator.py#L25

1. The dostring states "This method returns a result which allows access to the entire wave function". But it doesn't really. The returned `SimulationTrialResult` hides the final state in a private `_final_simulator_state` variable.

2. The array returned in `SimulationTrialResult._final_simulator_state` seems to be too large by a factor of 2, as if a qubit is being added somewhere.

3. The object in `SimulationTrialResult._final_simulator_state` is a numpy array, which causes `SimulationTrialResult.qubit_map` to fail.

4. The `initial_state` method parameter is silently ignored. It does does seem possible to set the initial state. 

5. The test `test_cirq_qsim_simulate_fullstate` does not test any properties of the final fullstate.

95-martin-orion:
Fixes #40: allows users to get final state from the result with `result.final_state`. The final state is an ordered vector of complex-valued amplitudes for each output state; this format should match the output of e.g. `WaveFunctionSimulator` in Cirq.

@gecrooks, please let me know if you see any issues with this PR.

### DIFF
--- a/qsimcirq/qsim_simulator.py
+++ b/qsimcirq/qsim_simulator.py
@@ -14,7 +14,16 @@
 
 from typing import Any, Dict, List, Sequence
 
-from cirq import study, ops, circuits, protocols, SimulatesAmplitudes, SimulatesFinalState, SimulationTrialResult
+from cirq import (
+  circuits,
+  ops,
+  protocols,
+  sim,
+  study,
+  SimulatesAmplitudes,
+  SimulatesFinalState,
+  SimulationTrialResult,
+)
 
 import numpy as np
 
@@ -22,22 +31,22 @@ from qsimcirq import qsim
 import qsimcirq.qsim_circuit as qsimc
 
 
-class QSimSimulatorState():
+class QSimSimulatorState(sim.WaveFunctionSimulatorState):
 
     def __init__(self,
                  qsim_data: np.ndarray,
                  qubit_map: Dict[ops.Qid, int]):
-      self.qubit_map = qubit_map
       # Generate state vector from qsim results
       amplitudes = qsim_data
-      N = len(self.qubit_map)
+      N = len(qubit_map)
       amplitudes = np.reshape(amplitudes, [2]*(N+1))
       amplitudes = np.transpose(amplitudes)
       amplitudes = amplitudes[0] + 1j * amplitudes[1]
-      self.state_vector = amplitudes.flatten()
+      state_vector = amplitudes.flatten()
+      super().__init__(state_vector=state_vector, qubit_map=qubit_map)
 
 
-class QSimSimulatorTrialResult(SimulationTrialResult):
+class QSimSimulatorTrialResult(sim.WaveFunctionTrialResult):
     
     def __init__(self,
                  params: study.ParamResolver,
@@ -46,9 +55,6 @@ class QSimSimulatorTrialResult(SimulationTrialResult):
       super().__init__(params=params,
                        measurements=measurements,
                        final_simulator_state=final_simulator_state)
-
-      # This field exposes the final state to external users.
-      self.final_state = final_simulator_state.state_vector
 
 
 class QSimSimulator(SimulatesAmplitudes, SimulatesFinalState):
--- a/qsimcirq/qsim_simulator.py
+++ b/qsimcirq/qsim_simulator.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Any, List, Sequence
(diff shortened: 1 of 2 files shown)


qsimcirq/qsim_simulator.py
```diff
@@ -22,6 +22,35 @@
...
+      # Generate state vector from qsim results
+      amplitudes = qsim_data
+      N = len(self.qubit_map)
+      amplitudes = np.reshape(amplitudes, [2]*(N+1))
+      amplitudes = np.transpose(amplitudes)
+      amplitudes = amplitudes[0] + 1j * amplitudes[1]
+      self.state_vector = amplitudes.flatten()
+
+
+class QSimSimulatorTrialResult(SimulationTrialResult):
```
gecrooks: I think you want to inherit from WaveFunctionTrialResult here. It's what cirq.Simulator() returns and it takes care of setting the state_vector, plus some other convenience methods.

95-martin-orion: Good point. Did similarly for the `QSimSimulatorState` object.

qsimcirq_tests/qsimcirq_test.py
```diff
@@ -53,18 +53,38 @@ def test_cirq_qsim_simulate_fullstate(self):
...
+    # Expected output state is:
+    # |1> (|01> + |10>) (|0> - |1>)
+    # = 1/2 * (|1010> - |1011> + |1100> - |1101>)
 
     qsim_circuit = qsimcirq.QSimCircuit(cirq_circuit)
 
     qsimSim = qsimcirq.QSimSimulator()
-    result = qsimSim.simulate(qsim_circuit)
+    result = qsimSim.simulate(qsim_circuit, qubit_order=[a, b, c, d])
+    print(result)
```
gecrooks: Delete unneeded print?

qsimcirq_tests/qsimcirq_test.py
```diff
@@ -53,18 +53,38 @@ def test_cirq_qsim_simulate_fullstate(self):
...
+    # Create a circuit.
     cirq_circuit = cirq.Circuit(
-        cirq.X(a)**0.5,  # Square root of X.
-        cirq.Y(b)**0.5,  # Square root of Y.
-        cirq.Z(c),  # Z.
-        cirq.CZ(a, d)  # ControlZ.
+        cirq.Moment([
+            cirq.X(a)**0.5,  # Square root of X.
+            cirq.H(b),       # Hadamard.
+            cirq.X(c),       # X.
```
gecrooks: I think you should add a Y gate here. At present the resulting wavefunction has no complex entries, so you're not really testing that the float->complex conversion is working.

qsimcirq_tests/qsimcirq_test.py
```diff
@@ -53,18 +53,38 @@ def test_cirq_qsim_simulate_fullstate(self):
...
+    # = 1/2 * (|1010> - |1011> + |1100> - |1101>)
 
     qsim_circuit = qsimcirq.QSimCircuit(cirq_circuit)
 
     qsimSim = qsimcirq.QSimSimulator()
-    result = qsimSim.simulate(qsim_circuit)
+    result = qsimSim.simulate(qsim_circuit, qubit_order=[a, b, c, d])
+    print(result)
+    assert result.final_state.shape == (16,)
+    np.testing.assert_array_almost_equal(
```
gecrooks: Suggest comparing against cirq's simulator. Something like
`cirq_final_state = cirq.Simulator(cirq_circuit).state_vector()`

gecrooks: Apart from those minor suggestions,  LGTM. Generating correct results in my code.

qsimcirq_tests/qsimcirq_test.py
95-martin-orion: Currently trying this with an S gate (to ensure that the complex component can't be extracted to the global phase) but I'm seeing issues. Might be related to #44.

95-martin-orion: Done. Part of the issue was a global phase applied in qsim but not in Cirq; I've accounted for this in the test.

qsimcirq_tests/qsimcirq_test.py
95-martin-orion: Done.

PR #45, opened 2020-03-12
Fix iSwap gate matrix

gecrooks:


95-martin-orion:
This seems correct to me, but I'll defer to Sergei for confirmation. We'll need to update tests before submitting.

gecrooks:
The problem is that practically everyone else defines the iSwap gate as the complex conjugate of what qsim has. This is bound to cause confusion. In particular, cirq has the normal convention, and thus the conversion of iSwap from cirq to qsim doesn't currently work.

https://github.com/quantumlib/qsim/blob/8d7f4436d0e35253eac74682e75e002650d31182/qsimcirq/qsim_circuit.py#L109

sergeisakov:
I think that in the quantum supremacy paper we used the convention iSwap = fSim(pi/2, 0). I don't mind using the standard convention but please update the tests and Schmidt decomposition matrices.

gecrooks:
Yeah, fSim(pi/2, 0) is the conjugate of iSwap (although its not a big deal because iSwap and its inverse are locally related by 1-qubit gates).

I'll take a look at the tests, but it's not immediately obvious to me how to modify the Schmidt decomposition matrices. What's the encoding?

sergeisakov:
Currently, the Schmidt decomposition is iSwap = (sqrt(1/2) I) x (sqrt(1/2) I) + (sqrt(-i/2) X) x (sqrt(-/2) X) + (sqrt(-i/2) Y) x (sqrt(-i/2) Y) + (sqrt(1/2) X) x (sqrt(1/2) Z). Note that the Schmidt coefficients are absorbed by the matrices So (sqrt(-i/2) X) and (sqrt(-i/2) Y) should be replaced by (sqrt(i/2) X) and (sqrt(i/2) Y) in GateIS in gates_def.h (X: schmidt_decomp[1][0], schmidt_decomp[1][1] and Y: schmidt_decomp[2][0], schmidt_decomp[2][1]).

PR #46, opened 2020-03-12
Add helper function to run qsim

sergeisakov:


tests/run_qsim_test.cc
```diff
@@ -62,7 +62,7 @@ R"(4
 10 h 3
 )";
 
-TEST(QSimRunner, RimQSim) {
+TEST(QSimRunner, RimQSim1) {
```
95-martin-orion: Should this be `RunQSim1`? Similarly for the other test.

sergeisakov: Yes, thank you!

PR #50, opened 2020-03-13
Refactor gate interface.

sergeisakov:


docs/type_reference.md
```diff
@@ -19,7 +19,9 @@ they fulfill the same expectations.
 | Function (parfor.h)     | Any function; args are specified with `Args`.     |
 | FuserT                  | [`BasicGateFuser`](lib/fuser_basic.h)             |
 | Gate                    | [`Gate`](lib/gate.h)                              |
+| GateQSim                | [`GateQSim`](lib/gates_def.h)                     |
 | GateDef                 | [`GateRX`](lib/gatedef.h)                         |
+| GK / GateKind           | [`Gate`](lib/gate.h)
```
95-martin-orion: This link should probably be [`GateKind`](lib/gate.h) to distinguish from the `Gate` type above.

lib/gates_def.h
```diff
@@ -26,6 +26,33 @@
...
+  kGateCZ,      // CZ
+  kGateCNot,    // CNOT (CX)
+  kGateIS,      // iSwap
+  kGateFS,      // fSim
+  kGateCP,      // control phase
+  kGateDecomp,  // single qubit gate from Schmidt decomposition
+};
+
+template <typename fp_type>
+using GateQSim = Gate<fp_type, GateKind>;
```
95-martin-orion: Personally, `GateQSim` doesn't help me distinguish between this and the base `Gate` type (aren't all gates qsim gates?) Maybe `DefaultGate` or similar would be clearer, what do you think?

docs/type_reference.md
sergeisakov: Fixed.

lib/gates_def.h
sergeisakov: Basically, GateQSim refers to the gates implemented in qsim. Gate is a generic gate. The idea is to make it easier to use qsim with external gate sets, which differ from the gate set implemented in qsim.

95-martin-orion: Okay, in the context of external gate sets that naming makes sense. To clarify this meaning for users, could you add a comment in the code here?

sergeisakov: Done.

Issue #48, opened 2020-03-13
Feature request: SWAP gate


gecrooks:
This would facilitating running cirq circuits with qsim. In principle, with all-to-all connectivity, you can always remove the SWAP gates from the circuit and relabel qubits. But I don't think there is an easy way of doing that in cirq.

PR #52, opened 2020-03-13
Emulate SWAP with fSim.

95-martin-orion:
As requested in #48. Note that iSWAP can also be emulated with fSim but isn't, which suggests that either it was added first or that there's a benefit to explicitly implementing he more specific gate.

@gecrooks for input.


gecrooks:
No, you can't emulate SWAP with fSim. fSim(-π/2, 0) is iSwap, fSim(+π/2, 0) is the conjugate of iSwap, and fSim(-π/2, π/2)  is locally equivalent, but not identical, to SWAP.

Issue #57, opened 2020-03-16
Inconsistent versions


gecrooks:
The version string in setup.py is currently '0.0.1',

https://github.com/quantumlib/qsim/blob/cb02f33c9882da81ce09e1ae043b7fd5ed277c41/setup.py#L79

but pypi thinks the latest version is '0.0.2'

https://pypi.org/project/qsimcirq/0.0.2/

Also, it would be useful for qsimcirq to have a version attribute `__version__` for programatic access. This does not currently work...

```
> python -c 'import qsimcirq; print(qsimcirq.__version__)'
```

95-martin-orion:
While this issue is raised, we should also add a doc for how to perform the release process - including permissions required and any extra tests above and beyond the CI suite that should be run.

@iamvamsikrishnad, could you look into this? If I understand correctly, getting the versions to match will require a new release (i.e. v0.0.3).

gecrooks:
Another tangentially related issue is that qsim is currently locked to version 0.6 of cirq, but AFAICT everything works just fine with the latest 0.7 cirq.

https://github.com/quantumlib/qsim/blob/cb02f33c9882da81ce09e1ae043b7fd5ed277c41/requirements.txt#L3

iamvamsikrishnad:
> While this issue is raised, we should also add a doc for how to perform the release process - including permissions required and any extra tests above and beyond the CI suite that should be run.
> 
> @iamvamsikrishnad, could you look into this? If I understand correctly, getting the versions to match will require a new release (i.e. v0.0.3).

I created a seperate issue for the doc.

95-martin-orion:
These are now consistent at `0.0.2` - further work is tracked in #59.

PR #60, opened 2020-03-20
Fixversionnumbering #57

iamvamsikrishnad:


setup.py
```diff
@@ -74,9 +74,11 @@ def build_extension(self, ext):
 # README file as long_description.
 long_description = open('README.md', encoding='utf-8').read()
 
+__version__ = '0.0.2'
```
95-martin-orion: Will this behave correctly when pushing a new release? I thought v0.0.2 already existed.

iamvamsikrishnad: You are correct. When pushing a new release, this needs to be updated to 0.0.3. I changed it to 0.0.2 so that when a new version needs to be pushed it can be changed to 0.0.3.

Issue #39, opened 2020-03-07
New Features for qsim


MichaelBroughton:
Some nice features to have for qsim might be:
1. The ability to compute inner products between two state objects.
2. The ability to actually draw X number of bitstring samples from a state object.

95-martin-orion:
In particular, item (2) is high priority for the qsim-Cirq interface as well. Given that it's a requirement for both TFQ and Cirq, the implementation should live in the C++ code.

sergeisakov:
Cirq doesn't have the ability to sample from a state, does it? I have some code for sampling. I'll add it.

95-martin-orion:
Cirq does have code for sampling from a state, but for usability I think it makes sense for qsim to provide this behavior directly. Thanks for looking into it.

95-martin-orion:
> Cirq doesn't have the ability to sample from a state, does it? I have some code for sampling. I'll add it.

@sergeisakov, do you have an update on this? I'm starting to hit this requirement on non-TFQ integration work.

MichaelBroughton:
If you want you can copy the sampling code in TFQ to keep things moving and then improve it when a faster implementation comes along

https://github.com/tensorflow/quantum/blob/master/tensorflow_quantum/core/qsim/state_space.h#L55

Just a thought :)

95-martin-orion:
The features discussed here are now available in core qsim. #65 tracks adding explicit sampling support in qsimcirq.

PR #63, opened 2020-03-25
Add state sampler

95-martin-orion:
Adds a sampler for quantum states, (mostly) copied from TFQuantum. See #39.

Remaining work before review: add tests to match those in TFQ.

### DIFF
--- a/lib/state_sampler.h
+++ b/lib/state_sampler.h
@@ -12,47 +12,41 @@ class StateSampler {
  public:
   using State = typename StateSpace::State;
 
-  // Function to draw m samples from a StateSpace Object in
-  // O(2 ** num_qubits + m * log(m)) time.
-  // Samples are stored as bit encoded integers.
-  //
-  // This method takes a seed for the RNG. To choose a random seed,
-  // Use the other version of this method.
-  void SampleState(const StateSpace& statespace, const State& state,
-                   const int m, unsigned seed,
+  void SampleState(const StateSpace& state_space, const State& state,
+                   const unsigned num_samples, unsigned seed,
                    std::vector<uint64_t>* samples) {
-    if (m == 0) {
+    if (num_samples <= 0) {
       return;
     }
-    std::mt19937 mt(seed);
-    std::uniform_real_distribution<float> dist(0.0, 1.0);
+    samples->reserve(num_samples);
 
-    double cdf_so_far = 0.0;
-    std::vector<float> random_vals(m, 0.0);
-    samples->reserve(m);
-    for (int i = 0; i < m; i++) {
-      random_vals[i] = dist(mt);
+    double norm = 0;
+    for (uint64_t k = 0; k < state_space.Size(state); ++k) {
+      norm += std::norm(state_space.GetAmpl(state, k));
     }
-    std::sort(random_vals.begin(), random_vals.end());
 
-    int j = 0;
-    for (uint64_t i = 0; i < statespace.Size(state); i++) {
-      const std::complex<float> f_amp = statespace.GetAmpl(state, i);
-      const std::complex<double> d_amp = std::complex<double>(
-        static_cast<double>(f_amp.real()),
-        static_cast<double>(f_amp.imag()));
-      cdf_so_far += std::norm(d_amp);
-      while (random_vals[j] < cdf_so_far && j < m) {
-        samples->push_back(i);
-        j++;
+    std::mt19937_64 rgen(seed);
+    double rgen_norm = norm / std::mt19937_64::max();
+
+    std::vector<double> rs;
+    rs.reserve(num_samples);
+
+    for (unsigned i = 0; i < num_samples; ++i) {
+      rs.emplace_back(rgen() * rgen_norm);
+    }
+
+    std::sort(rs.begin(), rs.end());
+
+    unsigned m = 0;
+    double csum = 0;
+
+    for (uint64_t k = 0; k < state_space.Size(state); ++k) {
+      csum += std::norm(state_space.GetAmpl(state, k));
+
+      while (m < num_samples && rs[m] <= csum) {
+        samples->emplace_back(k);
+        ++m;
       }
-    }
-
-    // Safety measure in case of state norm underflow.
-    // Likely to not have huge impact.
-    while (j < m) {
-      samples->push_back(samples->at(samples->size() - 1));
-      j++;
     }
   }
 };
--- a/lib/state_sampler.h
+++ b/lib/state_sampler.h
@@ -12,13 +12,6 @@ class StateSampler {
  public:
   using State = typename StateSpace::State;
 
-  // Sample using a random seed.
-  void SampleState(const StateSpace& statespace, const State& state,
-                   const int m, std::vector<uint64_t>* samples) {
-    std::random_device device;
-    SampleState(statespace, state, m, device(), samples);
-  }
-
   // Function to draw m samples from a StateSpace Object in
   // O(2 ** num_qubits + m * log(m)) time.
   // Samples are stored as bit encoded integers.
--- a/lib/state_sampler.h
+++ b/lib/state_sampler.h
@@ -1,0 +1,59 @@
+#ifndef STATE_SAMPLER_H_
+#define STATE_SAMPLER_H_
+
+#include <complex>
+#include <memory>
+#include <random>
+
+namespace qsim {
+
+template <typename StateSpace>
+class StateSampler {
+ public:
+  using State = typename StateSpace::State;
+
+  // Function to draw m samples from a StateSpace Object in
+  // O(2 ** num_qubits + m * log(m)) time.
+  // Samples are stored as bit encoded integers.
+  void SampleState(const StateSpace statespace, const State& state, const int m,
+                std::vector<uint64_t>* samples) {
+    if (m == 0) {
+      return;
+    }
+    std::random_device device;
+    std::mt19937 mt(device());
+    std::uniform_real_distribution<float> dist(0.0, 1.0);
+
+    double cdf_so_far = 0.0;
+    std::vector<float> random_vals(m, 0.0);
+    samples->reserve(m);
+    for (int i = 0; i < m; i++) {
+      random_vals[i] = dist(mt);
+    }
+    std::sort(random_vals.begin(), random_vals.end());
+
+    int j = 0;
+    for (uint64_t i = 0; i < statespace.Size(state); i++) {
+      const std::complex<float> f_amp = statespace.GetAmpl(state, i);
+      const std::complex<double> d_amp = std::complex<double>(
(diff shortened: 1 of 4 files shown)


95-martin-orion:
> Remaining work before review: add tests to match those in TFQ.

Also, qsimcirq needs to provide hooks for this behavior.

95-martin-orion:
> Remaining work before review: add tests to match those in TFQ.

Done.

> Also, qsimcirq needs to provide hooks for this behavior.

Spun off #65 to address this - will likely require changes to the pybind module.

tests/state_sampler_test.cc
```diff
@@ -0,0 +1,136 @@
...
+
+#include "../lib/gates_appl.h"
+#include "../lib/gates_def.h"
+#include "../lib/parfor.h"
+#include "../lib/simulator_avx.h"
+#include "../lib/state_sampler.h"
+#include "../lib/statespace.h"
+
+namespace qsim {
+
```
MichaelBroughton: you might want to put an anonymous namespace in here so

```c++
namespace qsim {
namespace {

```

tests/state_sampler_test.cc
```diff
@@ -0,0 +1,136 @@
...
+    counts[val] += 1.0;
+  }
+  for (unsigned i = 0; i < counts.size(); ++i) {
+    EXPECT_NEAR(counts[i] / num_samples, expected_prob[i], 1e-2);
+  }
+}
+
+}  // namespace qsim
+
+int main(int argc, char** argv) {
```
MichaelBroughton: Do you need an int main here ? Can you not just bazel test <this file> ?

tests/state_sampler_test.cc
```diff
@@ -0,0 +1,136 @@
...
+  std::vector<uint64_t> measurements;
+  sampler.SampleState(state_space, state, num_samples, seed, &measurements);
+  ASSERT_EQ(measurements.size(), num_samples);
+  for (unsigned i = 0; i < measurements.size(); ++i) {
+    int val = measurements[i];
+    counts[val] += 1.0;
+  }
+  for (unsigned i = 0; i < counts.size(); ++i) {
+    EXPECT_NEAR(counts[i] / num_samples, expected_prob[i], 1e-2);
+  }
```
MichaelBroughton: I don't think you need such an elaborate mechanism here since you have no chance of sampling anything other than 1. Maybe you could just reduce this to:

```c++
for (int i =0;i<measurements.size();i++) {
    EXPECT_EQ(val, 1);
}
```
And reduce the samples down from 100,000 ?

tests/state_sampler_test.cc
```diff
@@ -0,0 +1,136 @@
...
+  auto state = state_space.CreateState();
+  state_space.SetStateZero(state);
+
+  unsigned num_samples = 100000;
+  unsigned seed = 0;
+
+  // Assign amplitudes for each state with a rotation in the complex plane.
+  std::vector<float> expected_prob = {0.0,  0.05, 0.07, 0.1,
+                                      0.25, 0.2,  0.18, 0.15};
+  float pi = 3.14159265358979323846;
```
MichaelBroughton: Could probably use the globally defined M_PI symbol here instead ?

95-martin-orion:
> I think that this state sampler is biased because the state norm doesn't sum to one.

Why is this the case? As I understand it, `std::norm(ampl)` returns `|ampl|^2`. For a superposition of states with amplitudes `ampl[]`, I would expect the norms of each to sum to one.

lib/state_sampler.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace qsim {
+
+template <typename StateSpace>
+class StateSampler {
+ public:
+  using State = typename StateSpace::State;
+
+  // Sample using a random seed.
+  void SampleState(const StateSpace statespace, const State& state,
```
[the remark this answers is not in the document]
95-martin-orion: Done.

tests/state_sampler_test.cc
95-martin-orion: Good point. `num_samples` at 100k doesn't really increase test runtime, but I dropped it to 1k just in case.

tests/state_sampler_test.cc
95-martin-orion: This is to support `make run-cxx-tests`. If Make becomes cumbersome to support we might consider dropping it in the future, but for now I think it's still reasonable to keep around.

lib/state_sampler.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace qsim {
+
+template <typename StateSpace>
+class StateSampler {
+ public:
+  using State = typename StateSpace::State;
+
+  // Sample using a random seed.
+  void SampleState(const StateSpace& statespace, const State& state,
```
sergeisakov: I think such a function shouldn't be provided as it leads to unreproducible results.

95-martin-orion:
As a preface: one upside to the current sampling method is that it doesn't need to iterate over all states (although it does so currently). This can greatly reduce sampling runtime for e.g. Porter-Thomas distributions over many-qubit circuits.

> The reason is that sum_i |c_i|^2 != 1. For instance, if sum_i |c_i|^2 < 1 then the largest bitstring will be overrepresented.

What are the sources and scale of error here? If it's only floating-point underflow, I don't think we need special handling for the sum; even for large numbers of qubits such effects should be negligible (double has 52 data bits -> max underflow is of the order `2^(qubits-52)`, or 2^-15 ~= 1/32,000 for 37 qubits)

> I have been working on a state sampler today {...}
> ```
>     double norm = 0;
>     for (uint64_t k = 0; k < state_space.Size(state); ++k) {
>       norm += std::norm(state_space.GetAmpl(state, k));
>     }
> ```

As noted above, this requires iterating over all states, which could take a while for many-qubit circuits.

sergeisakov:
How do you propose to iterate over only a fraction of states? One can just exit the for loop when j == m. However, this greatly reduces sampling runtime only when m is small (m < 10).

There are basically two sources of error: single precision and round-off errors. The effect is quite severe. For instance, I have 1 - sum_i |c_i|^2 = 2.89e-06 for circuits/circuit_q30 of depth 8. So if you need, say, 1e7 samples then the sampling that uses the tfq code will be significantly biased towards some particular bitstring (the last one that gets picked up in the for loop).

Iterating over all states is typically quite fast compared to circuit simulation when using the same number of threads. We iterate over all states many times to simulate a large circuit.

MichaelBroughton:
For the record that is 30 bitstrings that would come up out of 10 million. If you are concerned about the last bitstring having a probability of almost 0 and then having 30 extra samples of it showing up, maybe instead we can just pick the index in the wavefunction that has the highest probability and then add 30 more samples to that entry in the case of norm underflow. What do you think ?

95-martin-orion:
> How do you propose to iterate over only a fraction of states? One can just exit the for loop when j == m. However, this greatly reduces sampling runtime only when m is small (m < 10).

Disregard my previous comment; I had made an assumption (that the states were ordered by probability) that is incorrect.

95-martin-orion:
Given that iterating over all elements is unavoidable, it seems reasonable to me to use Sergei's approach - it won't significantly impact runtime, and it gives us more confidence in the resulting samples. I'll send another commit with the fix.

MichaelBroughton:
That is true. Although when the states are this large, iterating over them once vs twice could lead to performance differences.

95-martin-orion:
> That is true. Although when the states are this large, iterating over them once vs twice could lead to performance differences.

Manual tests on a 30-qubit, depth-8 circuit using 8 threads show a noticeable difference in sampling runtime with the calculated sum-of-norms: using the current code, sampling takes ~10% the time of the simulation; with calculated sum-of-norms, it's closer to ~20%. (Note however that these are rough measures, and that the impact of sampling is less pronounced for larger circuit depth)

There are ways around this; of note is the use of precomputed sums, as described [here](https://stackoverflow.com/questions/6052603/fast-weighted-random-selection-from-very-large-set-of-values/6641764). This would make the initial pass O(n), with subsequent passes taking O(n/m) for `m` stored sums.

sergeisakov:
> For the record that is 30 bitstrings that would come up out of 10 million. If you are concerned about the last bitstring having a probability of almost 0 and then having 30 extra samples of it showing up, maybe instead we can just pick the index in the wavefunction that has the highest probability and then add 30 more samples to that entry in the case of norm underflow. This would move the bias towards the highest probability bitstring instead. It would also allow sampling to happen in one pass over the wavefunction instead of two.
> 
> What do people think ?

I think it doesn't fix the bias. Though this bias (with or without your fix) should have negligible effect for many applications.

sergeisakov:
> Manual tests on a 30-qubit, depth-8 circuit using 8 threads show a noticeable difference in sampling runtime with the calculated sum-of-norms:

You should use one thread because sampling is not paralellized.

> There are ways around this; of note is the use of precomputed sums, as described [here](https://stackoverflow.com/questions/6052603/fast-weighted-random-selection-from-very-large-set-of-values/6641764). This would make the initial pass O(n), with subsequent passes taking O(n/m) for `m` stored sums.

I doubt it will be much faster.

95-martin-orion:
Given that the runtime of sampling is O(n) in either case, I think avoiding bias should take priority here. I recommend that we submit Sergei's implementation (now included in this PR) and push runtime improvements to a future PR.

lib/state_sampler.h
```diff
@@ -0,0 +1,56 @@
...
+    }
+    samples->reserve(num_samples);
+
+    double norm = 0;
+    for (uint64_t k = 0; k < state_space.Size(state); ++k) {
+      norm += std::norm(state_space.GetAmpl(state, k));
+    }
+
+    std::mt19937_64 rgen(seed);
+    double rgen_norm = norm / std::mt19937_64::max();
```
sergeisakov: It is probably better to use `std::uniform_real_distribution(0.0, norm)` as in your initial commit.

lib/state_sampler.h
```diff
@@ -0,0 +1,56 @@
...
+
+    std::sort(rs.begin(), rs.end());
+
+    unsigned m = 0;
+    double csum = 0;
+
+    for (uint64_t k = 0; k < state_space.Size(state); ++k) {
+      csum += std::norm(state_space.GetAmpl(state, k));
+
+      while (m < num_samples && rs[m] <= csum) {
```
sergeisakov: This can be replaced by `while (rs[m] <= csum && m < num_samples)` if you replace `rs.reserve(num_samples)` by `rs.reserve(num_samples + 1)` above. Note that there is a bug in the original TFQ code: array access outside of its boundary. If you use `std::uniform_real_distribution` then replace `rs[m] <= csum` by `rs[m] < csum`.

95-martin-orion:
Closing this PR in favor of #68.

Issue #24, opened 2020-02-26
Move pybind requirement to the user


95-martin-orion:
As discussed offline, installing pybind11 as part of the `pip install qsimcirq` process may be driving the installation time way up. This is especially true for the Docker install test, where the Docker image must install `git` in order to install qsimcirq.

To remedy this, we should instead require that users install pybind11 on their system and add documentation explaining this requirement.

95-martin-orion:
`pip install qsimcirq` will skip pybind installation if the user already has it installed. Including pybind11 in the qsim requirements was the correct choice.

PR #66, opened 2020-03-27
Move pybind requirement to the user #24

iamvamsikrishnad:


95-martin-orion:
Closing this as too far separated from modern qsim. If we need to do this, we'll open a new PR for it.

PR #68, opened 2020-03-27
Add state sampler.

sergeisakov:


95-martin-orion:
To read from file in the Bazel test, you'll need to include `circuit_q24` as a data dependency in the BUILD rule. See qsim_base in the `apps` BUILD file for an example of this.

lib/statespace_avx.h
```diff
@@ -111,6 +112,57 @@ struct StateSpaceAVX final : public StateSpace<ParallelFor, float> {
         Base::num_threads_, Base::raw_size_ / 16, f, Op(), state);
   }
 
+  std::vector<uint64_t> Sample(
```
95-martin-orion: Does this make use of any AVX instructions? If not, I think we should provide a shared sampler for all statespace implementations.

95-martin-orion: My concern here isn't different floating-point types (although that's also a valid issue) - it's that statespace_basic has no equivalent sampling method.

sergeisakov: No, this doesn't make use of any AVX instructions. However, it does make use of specific state vector layout that is relevant only for the AVX version.

sergeisakov: I will add a sampler to statespace_basic in the next PR.

PR #71, opened 2020-03-31
Add state sampler.

sergeisakov:


tests/statespace_basic_test.cc
```diff
@@ -0,0 +1,38 @@
...
+
+#include "gtest/gtest.h"
+
+#include "../lib/parfor.h"
+#include "../lib/simulator_basic.h"
+#include "../lib/statespace_basic.h"
+
+namespace qsim {
+
+TEST(StateSpaceAVXTest, SamplingBasic) {
```
95-martin-orion: `TEST(StateSpaceBasicTest, ...)` instead? Same below.

tests/BUILD
```diff
@@ -88,10 +88,30 @@ cc_test(
     ],
 )
 
+cc_library(
```
95-martin-orion: I would recommend adding `testonly = 1,` to this rule (see [bazel docs](https://docs.bazel.build/versions/master/be/common-definitions.html) for testonly usage).

tests/statespace_test.h
```diff
@@ -0,0 +1,433 @@
...
+#include "../lib/circuit_reader.h"
+#include "../lib/fuser_basic.h"
+#include "../lib/gates_def.h"
+#include "../lib/io.h"
+#include "../lib/run_qsim.h"
+
+namespace qsim {
+
+template <typename StateSpace>
+void TestSamplingBasic() {
```
95-martin-orion: Minor nit: `TestSamplingBasic` suggests to me that this is for `StateSpaceBasic` only - maybe `TestSamplingSmall`?

sergeisakov:
Done. Thanks for your review!

PR #73, opened 2020-04-01
Add the ability to compute inner products.

sergeisakov:


lib/statespace_basic.h
```diff
@@ -94,6 +94,40 @@ struct StateSpaceBasic final : public StateSpace<ParallelFor, FP> {
...
+      const State& state1, const State& state2) const {
+    using Op = std::plus<std::complex<double>>;
+
+    auto f = [](unsigned n, unsigned m, uint64_t i, const State& state1,
+              const State& state2) -> std::complex<double> {
+      auto s1 = state1.get() + 2 * i;
+      auto s2 = state2.get() + 2 * i;
+
+      double re = s1[0] * s2[0] + s1[1] * s2[1];
+      double im = s1[0] * s2[1] - s1[1] * s2[0];
```
95-martin-orion, line 107: This seems reversed to me - shouldn't the real component have a subtraction (i.e. `re1*re2 - im1*im2`) and the imaginary component be addition only?

tests/statespace_test.h
```diff
@@ -387,14 +351,92 @@ R"(20
...
+  using StateSpace = typename Simulator::StateSpace;
+  using State = typename StateSpace::State;
+  using Runner = QSimRunner<IO, BasicGateFuser<GateQSim<float>>, Simulator>;
+
+  StateSpace state_space(circuit.num_qubits, 1);
+  State state0 = state_space.CreateState();
+
+  EXPECT_EQ(state_space.IsNull(state0), false);
+
+  auto measure = [&state0](unsigned k,
```
95-martin-orion, line 372: As I see it, this is taking the states at time `depth` and time `depth + 1` and checking their inner product - is that correct?

tests/statespace_test.h
```diff
@@ -387,14 +351,92 @@ R"(20
 30 y_1_2 19
 )";
 
+template <typename Simulator>
+void TestNormAndInnerProduct() {
```
95-martin-orion, line 355: As with TestSamplingSmall, it might help to have a test for the inner product methods that is verifiable by hand.

lib/statespace_basic.h
sergeisakov: The fist state should be complex conjugate. So this is reversed.

tests/statespace_test.h
sergeisakov: Yes, that is correct. There is only one T gate at time `depth + 1`. So the inner product can be estimated analytically: {(1 + cos(pi/4))/2, sin(pi/4)/2}.

tests/statespace_test.h
sergeisakov: Done.

PR #80, opened 2020-04-17
Short circuit before index-out-of-bounds.

95-martin-orion:
Fixes an address-sanitizer complaint - thanks to @MichaelBroughton for pointing this out. ASAN tests to be added at a later time.

This may also need to be applied to the SSE sampler.


95-martin-orion:
> Note that there is no out of boundary memory access as I allocate one extra array element in `GenerateRandomValues` in lib/util.h.

In that case, I think the correct solution is to populate that extra array element in `GenerateRandomValues`. Otherwise this will fail adress-sanitizer checks required for aligning with TFQ.

PR #81, opened 2020-04-20
Populate "extra" array element.

95-martin-orion:
This is required to prevent sanitizer errors. (The alternative, seen in #80, has a larger impact on runtime).


sergeisakov:
One thing is that TFQ had a real out of boundary error in the sampling code for a long time. Somehow address sanitizers didn't complain.

PR #82, opened 2020-04-21
Add simulator selector.

sergeisakov:


95-martin-orion:
The Docker test failure can be ignored for now - there was a [breaking change](https://github.com/googleapis/google-api-python-client/issues/876) in `google-api-python-client` that should be fixed in the next release. (Not clear how long until that release, though)

MichaelBroughton:
You can pin the version to 1.8.0 in your requirements files to fix this pretty easily. Should we just do that ?

95-martin-orion:
> You can pin the version to 1.8.0 in your requirements files to fix this pretty easily. Should we just do that ?

I agree with this solution - it would be more stable in the long run to have a pinned version and update manually.

busunkim96:
Hi! Dropping by since I saw you referenced https://github.com/googleapis/google-api-python-client/issues/876.

google-api-python-client 1.8.2 has been released and removes the problematic `googleapiclient.__version__` line. In addition, if you have any references to `apiclient` we recommend you replace them with `googleapiclient` instead.

PR #89, opened 2020-04-30
Refactor gate code.

sergeisakov:


sergeisakov:
Yes, that is correct. Thank you for the review.

PR #95, opened 2020-05-01
Split io.h.

sergeisakov:
Get a separate struct for file operations.


sergeisakov:
IO is for logging. IOFile is for both logging and file management.

PR #96, opened 2020-05-04
Restore "gates_qsim.h" include in hybrid.h

95-martin-orion:
`hybrid.h` still depends on some objects from `gate.h` and `gates_qsim.h`:
- `schmidt_decomp_type`
- `kGateDecomp`

I'm a little confused how our tests passed without this...we should look into this.

### DIFF
--- a/lib/hybrid.h
+++ b/lib/hybrid.h
@@ -16,12 +16,13 @@
 #define HYBRID_H_
 
 #include <algorithm>
+#include <array>
 #include <complex>
 #include <vector>
 
 #include "fuser_basic.h"
+#include "gate.h"
 #include "gates_appl.h"
-#include "gates_qsim.h"
 #include "util.h"
 
 namespace qsim {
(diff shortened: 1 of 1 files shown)


sergeisakov:
Thank you for catching this. I think it should be fixed in a different way. Please include `gate.h` instead of `gates_qsim.h` (please also include `<array>`). Add `using GateKind = typename Gate::GateKind;` below `using Gate = GateT;` and replace `kGateDecomp` by `GateKind::kGateDecomp`. `schmidt_decomp_type` and `GetSchmidtDecomp` are defined in `"gate.h"`.
The reason why our tests passed is that `"gates_qsim.h"` is included before `"hybrid.h"`.

95-martin-orion:
> Thank you for catching this. I think it should be fixed in a different way. {...}

Applied changes.

> The reason why our tests passed is that `"gates_qsim.h"` is included before `"hybrid.h"`.

Hmm...the Google-internal copies of the tests failed (which is how I caught this), but there are enough differences between those and the repo tests that this seems like the plausible culprit.

PR #99, opened 2020-05-05
Remove some compiler warnings.

sergeisakov:


lib/simulator_avx.h
```diff
@@ -153,6 +153,7 @@ class SimulatorAVX final {
       break;
     default:
       // Cannot reach here.
+      ml = _mm256_set_epi32(0, 0, 0, 0, 0, 0, 0, 0);
```
95-martin-orion, line 156: It's probably preferable to raise an error here, if we have a way to do so.

(Similarly for other "unreachable" sections in this file)

lib/statespace.h
```diff
@@ -48,8 +48,11 @@ class StateSpace {
       return State((fp_type*) _aligned_alloc(vector_size, 64), &_aligned_free);
     #else
       void* p = nullptr;
-      posix_memalign(&p, 64, vector_size);
-      return State((fp_type*) p, &free);
+      if (posix_memalign(&p, 64, vector_size) == 0) {
+        return State((fp_type*) p, &free);
+      } else {
+        return State(nullptr, &free);
```
95-martin-orion, line 54: Should this be an error, or is it a valid State object?

lib/simulator_avx.h
sergeisakov: How do you propose to raise an error here? These sections are really unreachable if those functions are used correctly. The correct usage is described in function comments.

lib/statespace.h
sergeisakov: No, it is an invalid State object. The validity can be tested by the `IsNull()` function as it is done in `run_qsim.h` and `hybrid.h`.

lib/simulator_avx.h
95-martin-orion: Sorry, I was in a Python state of mind. This should be fine as-is.

lib/statespace.h
95-martin-orion: Okay. As long as it is caught somewhere, this is fine.

PR #100, opened 2020-05-06
Fix setting a uniform state

sergeisakov:
This PR fixes SetStateUniform functions for one- and two-qubit circuits and a typo in statespace_basic.h.

### DIFF
--- a/tests/statespace_testfixture.h
+++ b/tests/statespace_testfixture.h
@@ -352,6 +352,35 @@ R"(20
 )";
 
 template <typename StateSpace>
+void TestNormSmall() {
+  using State = typename StateSpace::State;
+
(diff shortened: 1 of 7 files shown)


tests/statespace_testfixture.h
```diff
@@ -351,6 +351,35 @@ R"(20
 30 y_1_2 19
 )";
 
+template <typename StateSpace>
+void TestNormSmall() {
```
95-martin-orion, line 355: Does this test differentiate between the old and new behavior? i.e. will it fail if run without the other changes in this PR?

sergeisakov: Yes, it will fail without the other changes in this PR.

PR #107, opened 2020-05-13
Revert "Add rules to lib/BUILD."

95-martin-orion:
My request for aggregate libraries on PR #98 was in error - while the open-source version of bazel permits inclusion of header files from transitive `deps` fields, the Google-internal version does not. Reverting this commit is the cleanest way to bring the two back into agreement.


95-martin-orion:
> Would it be possible just to remove the run_qsim_lib and run_qsimh_lib rules from lib/BUILD and leave the rest (including the qsim_lib rule and comments)? Or even better, could you replace dependencies in the run_qsim_lib and run_qsimh_lib rules with the required header files?

This is a good point. I'll close this and send out a separate PR with these changes shortly.

PR #111, opened 2020-05-15
Remove extra linking requirements.

95-martin-orion:
Prior to this PR, additional static links `-lm` and `-lrt` were required to build qsim in the TFQ-compatible environment. This PR should remove these requirements.

### DIFF
--- a/apps/qsim_von_neumann.cc
+++ b/apps/qsim_von_neumann.cc
@@ -15,7 +15,6 @@
 #include <unistd.h>
 
 #include <algorithm>
-#include <cmath>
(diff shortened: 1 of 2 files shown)


95-martin-orion:
Note that the "Kokoro" build is expected to fail until fully set up. It should not block merging.

95-martin-orion:
Update: qsim_von_neumann is not the only binary that requires `-lm`, so I've restored `<cmath>`. If it proves infeasible to remove the `-lm` requirement, we can enforce it in the qsim-TFQ boundary.

PR #112, opened 2020-05-19
Fix hybrid simulator.

sergeisakov:
This PR fixes the hybrid simulator for gates with Schmidt rank 1 and 3. This PR also (slightly) changes the QSimHRunner interface.

### DIFF
--- a/lib/hybrid.h
+++ b/lib/hybrid.h
@@ -508,18 +508,6 @@ struct HybridSimulator final {
       // Not supported.
       return 42;
     }
-/*
-    if (size == 2) {
-      return 1;
-    } else if (size == 4 || size == 3) {
-      return 2;
-    } else if (size == 1) {
-      return 0;
-    }
-
-    // Not supported.
-    return 0;
-*/
   }
 
   static bool CreateStates(
--- a/lib/hybrid.h
+++ b/lib/hybrid.h
@@ -192,6 +192,12 @@ struct HybridSimulator final {
         }
 
         unsigned schmidt_bits = SchmidtBits(d.size());
+        if (schmidt_bits > 2) {
+          IO::errorf("Schmidt rank is too large for gate kind %u.\n",
+                     gate0.parent->kind);
+          return false;
+        }
+
         unsigned inverse = parts[gate0.parent->qubits[0]];
         if (gate0.parent->inverse) inverse = 1 - inverse;
         hd.gatexs.push_back(GateX{&gate0, nullptr, std::move(d), schmidt_bits,
@@ -203,6 +209,12 @@ struct HybridSimulator final {
     for (auto& gate1 : hd.gates1) {
       if (gate1.parent != nullptr) {
         hd.gatexs[count++].decomposed1 = &gate1;
+      }
+    }
+
+    for (auto& gatex : hd.gatexs) {
+      if (gatex.schmidt_decomp.size() == 1) {
+        FillSchmidtMatrices(0, gatex);
       }
     }
 
@@ -298,12 +310,16 @@ struct HybridSimulator final {
 
     std::vector<unsigned> prev(hd.num_gatexs, -1);
 
-    if (SetSchmidtMatrices(0, num_p_gates, param.prefix, prev, hd.gatexs)) {
+    unsigned gatex_index = SetSchmidtMatrices(
+        0, num_p_gates, param.prefix, prev, hd.gatexs);
+
+    if (gatex_index == 0) {
       // Apply gates before the first checkpoint.
       ApplyGates(fgates0, 0, loc0[0], sim0, state0p);
       ApplyGates(fgates1, 0, loc1[0], sim1, state1p);
     } else {
-      IO::errorf("invalid prefix.\n");
+      IO::errorf("invalid prefix %lu for prefix gate index %u.\n",
+                 param.prefix, gatex_index - 1);
       return false;
     }
 
@@ -314,7 +330,8 @@ struct HybridSimulator final {
         sspace1.CopyState(state1p, state1r);
       }
 
-      if (SetSchmidtMatrices(num_p_gates, num_pr_gates, r, prev, hd.gatexs)) {
+      if (SetSchmidtMatrices(num_p_gates, num_pr_gates,
+                             r, prev, hd.gatexs) == 0) {
         // Apply gates before the second checkpoint.
         ApplyGates(fgates0, loc0[0], loc0[1], sim0, state0r);
         ApplyGates(fgates1, loc1[0], loc1[1], sim1, state1r);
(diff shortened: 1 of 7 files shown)


lib/hybrid.h
```diff
@@ -419,7 +433,12 @@ struct HybridSimulator final {
       const auto& gatex = gatexs[i];
       unsigned k = (mask >> shift_length) & ((1 << gatex.schmidt_bits) - 1);
       shift_length += gatex.schmidt_bits;
+
       if (k != prev_k[i]) {
+        if (k >= gatex.schmidt_decomp.size()) {
+          return false;
```
95-martin-orion, line 439: From what I can tell, if this `return false` is hit then the system is in an error state, since it prevents any subsequent gates in `gatexs` from being used to update `prev_k`. For example:
```
Let gatexs.size() == 10.
Assume that this check fails at i == 2 and returns.
Then gatexs[3] to gatexs[9] are not checked.
Thus, prev_k[3] to prev_k[9] will not be correct.
```
Is this reading correct? How does this interact with the `continue`s above?

lib/run_qsimh.h
```diff
@@ -39,46 +39,55 @@ struct QSimHRunner final {
...
+    HybridData hd;
+    bool rc = HybridSimulator::SplitLattice(parts, circuit.gates, hd);
 
-    if (hd.num_gatexs < param.num_prefix_gatexs + param.num_root_gatexs) {
-      IO::errorf("error: num_prefix_gates (%u) plus num_root gates (%u) is "
-                 "greater than num_gates_on_the_cut (%u).\n",
-                 param.num_prefix_gatexs, param.num_root_gatexs, hd.num_gatexs);
-      return false;
-    }
+    if (rc) {
```
95-martin-orion, line 68: Style suggestion: put the short part of the conditional in the brackets to minimize indentation:
```
if (!rc) {
  return false;
}
// ...rest of function...
```

tests/hybrid_test.cc
```diff
@@ -73,7 +73,8 @@ R"(2
 
   std::vector<unsigned> parts = {0, 1};
 
-  auto hd = HybridSimulator::SplitLattice(parts, circuit.gates);
+  HybridSimulator::HybridData hd;
+  EXPECT_EQ(HybridSimulator::SplitLattice(parts, circuit.gates, hd), true);
```
95-martin-orion, line 77: This can use `EXPECT_TRUE`.

tests/hybrid_test.cc
```diff
@@ -73,7 +73,8 @@ R"(2
 
   std::vector<unsigned> parts = {0, 1};
 
-  auto hd = HybridSimulator::SplitLattice(parts, circuit.gates);
+  HybridSimulator::HybridData hd;
```
95-martin-orion, line 76: The new paths added in this PR should also have tests:
- Check that `SplitLattice` fails if `SetSchmidtMatrices` returns false before the first checkpoint
- Verify that `SplitLattice` recovers if `SetSchmidtMatrices` returns false after the first checkpoint (assuming this is the expected behavior)

lib/hybrid.h
```diff
@@ -292,11 +298,14 @@ struct HybridSimulator final {
...
-
-    // Apply gates before the first checkpoint.
-    ApplyGates(fgates0, 0, loc0[0], sim0, state0p);
-    ApplyGates(fgates1, 0, loc1[0], sim1, state1p);
+    if (SetSchmidtMatrices(0, num_p_gates, param.prefix, prev, hd.gatexs)) {
+      // Apply gates before the first checkpoint.
+      ApplyGates(fgates0, 0, loc0[0], sim0, state0p);
+      ApplyGates(fgates1, 0, loc1[0], sim1, state1p);
+    } else {
+      IO::errorf("invalid prefix.\n");
```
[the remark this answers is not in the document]
sergeisakov: Done. It's not the invalid gate. It's the invalid prefix that specifies the path.

lib/hybrid.h
```diff
@@ -305,11 +314,13 @@ struct HybridSimulator final {
...
-
-      // Apply gates before the second checkpoint.
-      ApplyGates(fgates0, loc0[0], loc0[1], sim0, state0r);
-      ApplyGates(fgates1, loc1[0], loc1[1], sim1, state1r);
+      if (SetSchmidtMatrices(num_p_gates, num_pr_gates, r, prev, hd.gatexs)) {
+        // Apply gates before the second checkpoint.
+        ApplyGates(fgates0, loc0[0], loc0[1], sim0, state0r);
+        ApplyGates(fgates1, loc1[0], loc1[1], sim1, state1r);
+      } else {
+        continue;
```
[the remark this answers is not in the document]
sergeisakov: We don't split the lattice here. We run simulation here. Basically, the `for (uint64_t r = 0; r < rmax; ++r) {` loop iterates over all the "root" paths. If a path is invalid then we reach this `continue`. Nothing bad happens here: we just continue to iterate over the rest of the paths. A path can be invalid only for gates with Schimdt rank 3 (in the first commit of this PR, it was also invalid for Schimdt rank 1 but I removed that (additions at the end of the `SplitLattice` function and `if (gatex.schmidt_bits == 0) {` in the `SetSchmidtMatrices` matrices)). We use 2 bits to encode the path for a Schimdt rank 3 gate. The bits `11` encode an invalid path.

lib/hybrid.h
sergeisakov: No, this reading is incorrect. We return false for an invalid path, see above. In this case, prev_k[3] to prev_k[9] will not be affected at all as we increment `mask` by 1 (`mask` encodes the path) and the path bits for gatexs[3] to gatexs[9] remain the same when the invalid path event is triggered for `i == 2`. There is a possibility to speed up things here by not iterating over the bits for `i < 2`. I will do this in some later PR as now we don't have gates with Schimdt rank 3 and it will not affect anything.

lib/run_qsimh.h
sergeisakov: Done.

tests/hybrid_test.cc
sergeisakov: Done.

tests/hybrid_test.cc
sergeisakov: `SetSchmidtMatrices` is not called in `SplitLattice`. Returning `false` (now it returns gate index plus one if the check fails) is okay for "root" and "suffix" paths. `Run` can fail for "prefix" paths. There is no way to recover as the prefix path is an input parameter. It cannot be tested because currently we don't have gates with Schimdt rank 3. Thank you for the review.

95-martin-orion: Sorry, I misread this on my first pass - I thought the calls to `SetSchmidtMatrices` in `Run` were in `SplitLattice` instead. This should be fine as-is then.

lib/hybrid.h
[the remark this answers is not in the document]
95-martin-orion: I see now - I think I missed the for-loop structure on my initial pass. This all looks good to me.

lib/hybrid.h
95-martin-orion: Sounds good to me.

PR #113, opened 2020-05-20
Refactor gates.

sergeisakov:


lib/gate.h
```diff
@@ -29,61 +28,35 @@ enum GateAnyKind {
 };
 
 // Gate is a generic gate to make it easier to use qsim with external gate sets.
-template <typename fp_type, typename GK = GateAnyKind>
+template <typename FP, typename GK = GateAnyKind>
```
95-martin-orion, line 31: I don't think this change adds much, especially with `using fp_type = FP;` immediately below.
(IIRC, `fp_type` is also the more commonly-used identifier for floating-point types in qsim)

sergeisakov: Thank you for the review. It is used in two functions below to ensure type safety.

PR #114, opened 2020-05-20
Add/update comments

sergeisakov:


lib/hybrid.h
```diff
@@ -460,6 +462,8 @@ struct HybridSimulator final {
 
       if (k != prev_k[i]) {
         if (k >= gatex.schmidt_decomp.size()) {
+          // Invalid path. Returns gatex index plus one to report error in case
```
95-martin-orion, line 465: More detail here could be useful - e.g., why does this indicate an invalid path?

lib/hybrid.h
```diff
@@ -452,6 +453,7 @@ struct HybridSimulator final {
       const auto& gatex = gatexs[i];
 
       if (gatex.schmidt_bits == 0) {
+        // Continue if gatex has Schmidt rank 1.
```
95-martin-orion, line 456: For clarity, maybe replace `Continue if...` with `Skip decomposition of gatex if...`

PR #116, opened 2020-05-21
Support Windows testing for Kokoro.

95-martin-orion:
Adds Windows-specific options to BUILD rules.

### DIFF
--- a/tests/BUILD
+++ b/tests/BUILD
@@ -26,7 +26,6 @@ windows_copts = [
     "/Iexternal/gemmlowp",
     "/wd4018",
     "/wd4577",
-    "/wd4267",  # Silence "narrowing conversion" errors.
     "/DNOGDI",
     "/UTF_COMPILE_LIBRARY",
 ]
--- a/tests/BUILD
+++ b/tests/BUILD
@@ -5,28 +5,8 @@ omp_copts = ['-fopenmp']
 omp_linkopts = ['-lgomp']
 
 windows_copts = [
-    "/D__CLANG_SUPPORT_DYN_ANNOTATION__",
-    "/D_USE_MATH_DEFINES",
-    "/DEIGEN_MAX_ALIGN_BYTES=64",
-    "/DEIGEN_HAS_TYPE_TRAITS=0",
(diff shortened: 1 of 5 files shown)


95-martin-orion:
Kokoro doesn't appear to trigger on draft PRs. Marking "ready for review" in an attempt to trigger Kokoro, but this PR is still a work in progress.

95-martin-orion:
> Kokoro doesn't appear to trigger on draft PRs. Marking "ready for review" in an attempt to trigger Kokoro, but this PR is still a work in progress.

This isn't necessarily true - I don't see Kokoro builds on _any_ PRs now - including ones where Kokoro previously ran! Something strange has happened. I'll look into this.

95-martin-orion:
> ...Something strange has happened. I'll look into this.

This appears to be tied to use of the `kokoro:run` label: only PRs with this label will trigger Kokoro. It's still not clear to me why old runs are no longer visible, but that's not too urgent at the moment.

95-martin-orion:
The Kokoro Windows build is now passing. Currently it only runs tests, as apps aren't a blocker for TFQ integration; for completeness I've added a TODO to also get the apps working in Windows.

It's likely _possible_ to build in Windows in the current state, but for TFQ integration there are some additional constraints that we're required to meet.

tests/BUILD
```diff
@@ -4,9 +4,23 @@ sse_copts = ['-msse4']
 omp_copts = ['-fopenmp']
 omp_linkopts = ['-lgomp']
 
+windows_copts = [
+    "/arch:AVX",
+    "/std:c++14",
```
sergeisakov, line 9: What is the reason to use C++14 instead of C++11?

95-martin-orion: C++14 is also used in the TFQ builds, which are what we need to align with for these tests. I'll double-check these with their team to be certain this was intended, but for now we'll stick with this.

(As an aside, and possibly a note to my future self: `/std:c++11` is not a recognized flag when building with Kokoro, as it uses C++11 by default.)

PR #125, opened 2020-06-05
Check all config flags in CI tests.

### LINKED ISSUE #124: Cover all flag combinations in tests
We currently support AVX, SSE, and unoptimized simulations, each of which can either use or exclude OpenMP. Our CI tests (and possibly our test definitions) need to be updated to cover all possible combinations of these tests.

95-martin-orion:
Fixes #124.

This ensures that all config flag combinations are tested in the CI tests, and allows a couple of tests (hybrid_test, run_qsim[h]_test) to run in SSE/unoptimized modes.

I think it may be possible to condense the simulator and statespace tests into a single file like this, but since all cases are currently covered I think we can consider that for a future PR.

### DIFF
--- a/.github/workflows/bazeltest.yml
+++ b/.github/workflows/bazeltest.yml
@@ -20,6 +20,12 @@ jobs:
   test:
     name: Test with Bazel
     runs-on: ubuntu-16.04
+    strategy:
+      matrix:
+        # Hardware optimizers.
+        hardware_opt: [avx,sse,basic]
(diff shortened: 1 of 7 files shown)


95-martin-orion:
> `Some checks haven’t completed yet`. Is that expected?

Yes, that's a side-effect of the old branch protection rules. I've updated them to the new checks, and it now reports "all checks passed".

Issue #134, opened 2020-06-19
Surface gates_cirq in qsimcirq interface


95-martin-orion:
There are currently a bunch of gates that core qsim supports in `gates_cirq.h` that are not being translated from Cirq. We need to surface this support so that users have access to it.

95-martin-orion:
Because the translation layer currently converts Cirq circuits to input files, this change forces us to make a decision:

1. Keep the current translation behavior, and add all new gates to the input file format.
2. Modify translation to produce qsim gate objects directly, and leave the input file format alone.

Option 1 is less work, but greatly increases the input file complexity and requires greater long-term effort to maintain the input file format.

Option 2 entails deprecation (but not removal) of the input file use case: future users will need to construct circuits in code or with Cirq to use gates outside the `gates_qsim` set. However, this is unavoidable for users who inject their own gate sets.

sboixo:
I think option 2 fits better with the philosophy that qsim is a library and cirq is the frontend to build circuits. Cirq can use the qsim library.

karlunho:
+1 on @sboixo suggestion

Issue #69, opened 2020-03-27
Consider full migration to Bazel


95-martin-orion:
There are a couple of discrepancies between the Bazel- and Makefile-driven builds of qsim - most recently, we found that input files referenced by Bazel require a different relative address than when building with Make.

One possible solution to this is to fully migrate off of Make and require all users to build qsim with Bazel. The goal of this issue is to collect feedback on this proposal; **we have not yet decided on this migration.**

PR #135, opened 2020-06-22
Translate directly from Cirq to C++ circuit.

95-martin-orion:
This should cover all of the gates required by #134. It also explicitly deprecates the file-input format; although we do not currently plan to remove file input, we will no longer maintain it. Python users are encouraged to use Cirq; C++ users will need to interface directly with the qsim libraries.

@sergeisakov: please confirm that this covers all of the `gates_cirq` gates (known exception: `MatrixGate`). If there are any other docs that need updating, let me know.

Follow-up work (will open new issues for these):
- MatrixGates
- Built-in decomposition of unknown gates


95-martin-orion:
Of note: I deleted `qsim_test.py` since that file was specifically for testing the "call qsim from Python without Cirq` use case, which is no longer supported. (Calling qsim directly from C++ is still valid, though)

qsimcirq/qsim_circuit.py
```diff
@@ -15,6 +15,82 @@
...
 
 import cirq
+from qsimcirq import qsim
+
+
+def _cirq_gate_kind(gate):
+  if isinstance(gate, cirq.ops.identity.IdentityGate):
+    if gate.num_qubits == 1:
+      return qsim.kI
+    return qsim.kI2
```
sergeisakov, line 25: Check if `if gate.num_qubits == 2` for `qsim.kI2`.

qsimcirq/qsim_circuit.py
```diff
@@ -15,6 +15,82 @@
...
+    return qsim.kPhasedISwapPowGate
+  if isinstance(gate, cirq.ops.FSimGate):
+    return qsim.kFSimGate
+  if isinstance(gate, cirq.ops.MatrixGate):
+    if gate.num_qubits == 1:
+      return qsim.kMatrixGate1
+    return qsim.kMatrixGate2
+  # No matches - decomposition is required.
+  # TODO: support decomposition.
+  return qsim.kGateDecomp
```
sergeisakov, line 93: `kGateDecomp` is a gate from Schmidt decomposition. It is internal to qsim. If there are no matches, an error should be raised.

pybind_interface/pybind_main.cpp
```diff
@@ -48,20 +48,12 @@ T parseOptions(const py::dict &options, const char *key) {
...
-  std::string circuit_str;
+Circuit<Cirq::GateCirq<float>> getCircuit(const py::dict &options) {
   try {
-    circuit_str = parseOptions<std::string>(options, "c\0");
+    return options["c\0"].cast<Circuit<Cirq::GateCirq<float>>>();
   } catch (const std::invalid_argument &exp) {
     throw;
   }
-  std::stringstream ss(circuit_str);
-  if (!CircuitQsimParser<IO>::FromStream(std::numeric_limits<unsigned>::max(),
```
sergeisakov, line 60: `#include "../lib/circuit_qsim_parser.h"` above should be deleted.

qsimcirq/qsim_circuit.py
95-martin-orion: Done.

qsimcirq/qsim_circuit.py
95-martin-orion: Done.

pybind_interface/pybind_main.cpp
95-martin-orion: Done.

95-martin-orion:
Regarding missing checks: I think Github's "required checks" and the "only execute on C++ changes" rule for Bazel aren't cooperating. Options include (1) making Bazel checks optional, or (2) running Bazel checks even when no C++ changes are submitted.

I'm not thrilled about it, but since (1) leaves holes in our test coverage I'll send out a PR for (2). (This might be good incentive for me to revisit issue #69...)

PR #136, opened 2020-06-22
Always run Bazel tests.

95-martin-orion:
As discussed in #135. Revisiting #69 (migrating all tests to Bazel) might make sense in this context.

### DIFF
--- a/.github/workflows/bazeltest.yml
+++ b/.github/workflows/bazeltest.yml
@@ -6,14 +6,7 @@ on:
     branches:
       - master
 
-  # Only run tests for PRs affecting C++ code.
   pull_request:
-    paths-ignore:
-    - '**.md'
-    - 'install/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/bazeltest.yml
```diff
@@ -6,14 +6,7 @@ on:
     branches:
       - master
 
-  # Only run tests for PRs affecting C++ code.
   pull_request:
-    paths-ignore:
-    - '**.md'
```
sergeisakov, line 12: Should `*.md` files still be ignored?

95-martin-orion: Sadly, Github expects "required checks" to run on _every_ PR, regardless of ignored paths. Ignoring `*.md` files here would prevent a docs-only change from being submitted.

PR #138, opened 2020-06-23
Support matrix gates and decomposition.

### LINKED ISSUE #137: Support MatrixGate and unknown gate decomposition in qsimcirq
Following #135, all `gates_cirq` gates are supported through qsimcirq except for matrix gates. These were excluded from the initial change because they require more complex translation logic, but going forward it would be nice to have support for them.

In a similar vein, there are some Cirq gates which are not in `gates_cirq`. This is basically unavoidable, as Cirq continues to add new gates. However, Cirq has a built-in `decompose` method to distill gates down to a target "gateset"; using this (with `gates_cirq` as the gateset) we can support any circuit sent from Cirq.

(Of note: `decompose` usually incurs a simulation cost compared to running the original gate. Having a fairly large target gateset like `gates_cirq` helps to minimize this cost.)

95-martin-orion:
Unrecognized gate types will now be decomposed to qsim-friendly types if possible, and matrix gates are now properly supported.

Fixes #137.

### DIFF
--- a/qsimcirq/qsim_circuit.py
+++ b/qsimcirq/qsim_circuit.py
@@ -96,8 +96,8 @@ def _cirq_gate_kind(gate):
     raise NotImplementedError(
       f'Received matrix on {gate.num_qubits()} qubits; '
       + 'only 1- or 2-qubit gates are supported.')
-  # TODO: support decomposing unrecognized gates.
-  raise NotImplementedError(f'Gate {gate} is of unrecognized type.')
+  # Unrecognized gates will be decomposed.
+  return None
 
 
 class QSimCircuit(cirq.Circuit):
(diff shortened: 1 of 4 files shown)


qsimcirq/qsim_circuit.py
```diff
@@ -149,15 +149,32 @@ def translate_cirq_to_qsim(
...
-          if isinstance(val, float)
-        }
-        qsim.add_gate(gate_kind, time, qubits, params, qsim_circuit)
+        qsim_ops = cirq.decompose(
+          op, keep=lambda x: _cirq_gate_kind(x.gate) != None)
+        moment_length = max(moment_length, len(qsim_ops))
+
+        for gi, qsim_op in enumerate(qsim_ops):
+          gate_kind = _cirq_gate_kind(qsim_op.gate)
+          time = mi + time_offset + gi
```
sergeisakov, line 162: Probably `time = time_offset + gi` is sufficient here.

95-martin-orion:
> Probably line 278 in `pybind_main.cpp` (`// TODO: support translating matrix gates.`) should be deleted.

Replaced with a comment on where matrix gates are handled (since this method still rejects them)

PR #139, opened 2020-06-25
Fix state order.

sergeisakov:


sergeisakov:
It was mostly motivated by possible incorrect behavior after switching to simmux. I also think that it's better to have this in qsim rather than in the Python interface.

Issue #74, opened 2020-04-02
Can gates in the circuit file be out of order?


95-martin-orion:
I'm working on a generator for qsim circuits, and for certain gates that qsim does not support it decomposes the gate into a sequence that qsim does support. If there are other gates running at the same time, this can create out-of-order gates in the file, e.g.:
```
# one input gate
0 rxy 0 0.5 0.25
1 rz 0 0.5
# another input gate
0 cp 1 2 0.75
```
Is this allowed in qsim or not? Either way, can we update `input_format.md` to clarify this for users?

(I have no preference for whether we *should* support this or not - if it's currently not allowed, I can refactor my generator code.)

sergeisakov:
It depends. Your example will work. However, the following examples will not work:
```
# one input gate
1 rz 0 0.5
0 rxy 0 0.5 0.25
# another input gate
0 cp 1 2 0.75
```
or
```
0 rxy 0 0.5 0.25
2 rz 0 0.5
0 cp 1 2 0.75
1 cz 0 1
```
That is gates acting on the same qubits must be in ascending order of time. This can be fixed by sorting gates or as you suggested, we can update `input_format.md` to clarify this.

PR #141, opened 2020-06-30
Add measurement gate.

sergeisakov:
This PR also cleans up statespace.

### DIFF
--- a/lib/fuser_basic.h
+++ b/lib/fuser_basic.h
@@ -24,7 +24,7 @@
 
 namespace qsim {
 
-template <typename Gate>
+template <typename IO, typename Gate>
 struct BasicGateFuser final {
   using GateFused = qsim::GateFused<Gate>;
 
@@ -103,8 +103,8 @@ struct BasicGateFuser final {
 
         if (gate.time < prev_time) {
           // This function assumes that gate times are ordered.
-          // Just stop sielently if this is not the case.
-          // TODO: report an error here.
+          // Just stop silently if this is not the case.
+          IO::errorf("gate times should be ordered.\n");
           gates_fused.resize(0);
           return gates_fused;
         }
--- a/lib/fuser_basic.h
+++ b/lib/fuser_basic.h
@@ -15,9 +15,11 @@
 #ifndef FUSER_BASIC_H_
 #define FUSER_BASIC_H_
 
+#include <map>
 #include <utility>
 #include <vector>
 
+#include "gate.h"
 #include "fuser.h"
 
 namespace qsim {
@@ -49,7 +51,7 @@ struct BasicGateFuser final {
    * applied together. Note that gates fused with this method are not
    * multiplied together until ApplyFusedGate is called on the output.
    * @param num_qubits The number of qubits acted on by 'gates'.
-   * @param gates The gates to be fused.
+   * @param gates The gates to be fused. Gate times should be ordered.
    * @param times_to_split_at Ordered list of time steps at which to separate
    *   fused gates. Each element of the output will contain gates from a single
    *   'window' in this list.
@@ -61,7 +63,17 @@ struct BasicGateFuser final {
       const std::vector<unsigned>& times_to_split_at) {
     std::vector<GateFused> gates_fused;
 
-    // Sequence of two-qubit gates and fixed single-qubit gates.
+    if (gates.size() == 0) return gates_fused;
+
+    gates_fused.reserve(gates.size());
+
+    // Merge with measurement gate times to separate fused gates at.
+    auto times = MergeWithMeasurementTimes(gates, times_to_split_at);
+
+    // Map to keep track of measurement gates with equal times.
+    std::map<unsigned, std::vector<const Gate*>> measurement_gates;
+
+    // Sequence of top level gates the other gates get fused to.
     std::vector<const Gate*> gates_seq;
 
     // Lattice of gates: qubits "hyperplane" and time direction.
@@ -70,7 +82,9 @@ struct BasicGateFuser final {
     // Current unfused gate index.
     std::size_t gate_index = 0;
 
-    for (std::size_t l = 0; l < times_to_split_at.size(); ++l) {
+    for (std::size_t l = 0; l < times.size(); ++l) {
+      if (gate_index == gates.size()) break;
+
       gates_seq.resize(0);
       gates_seq.reserve(gates.size());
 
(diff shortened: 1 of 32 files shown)


lib/statespace_avx.h
```diff
@@ -28,6 +28,31 @@
...
+  s1 = _mm256_and_si256(s1, ma);
+  s2 = _mm256_and_si256(s2, ma);
+
+  s1 = _mm256_cmpeq_epi64(s1, bi);
+  s2 = _mm256_cmpeq_epi64(s2, bi);
+
+  return _mm256_blend_epi32(s1, s2, 170);  // 10101010
+}
+
+inline double HorisontalSumAVX(__m256 s) {
```
95-martin-orion, line 48: `HorisontalSumAVX` -> `HorizontalSumAVX`?

lib/statespace.h
```diff
@@ -94,9 +104,67 @@ class StateSpace {
...
+    auto r = RandomValue(rgen, norm);
+
+    unsigned m = 0;
+    while (r > partial_norms[m]) ++m;
+    r -= m > 0 ? partial_norms[m - 1] : 0;
+
+    uint64_t k0 = For::GetIndex0(size, num_threads_, m);
+    uint64_t k1 = For::GetIndex1(size, num_threads_, m);
+
+    result.bits = FindMeasuredBits(k0, k1, r, result.mask, state);
```
95-martin-orion, line 153: The way this uses `For` indices and the structure of the `FindMeasuredBits` method itself give me the sense that this ought to use `For::Run` - could you comment on this?

lib/fuser_basic.h
```diff
@@ -79,13 +93,33 @@ struct BasicGateFuser final {
...
       for (; gate_index < gates.size(); ++gate_index) {
         const auto& gate = gates[gate_index];
 
-        if (gates[gate_index].time > times_to_split_at[l]) break;
+        if (gate.time > times[l]) break;
+
+        if (gate.time < prev_time) {
+          // This function assumes that gate times are ordered.
+          // Just stop sielently if this is not the case.
+          // TODO: report an error here.
```
95-martin-orion, line 107: `sielently` -> `silently`

Also: can the TODO be addressed in this PR?

lib/fuser_basic.h
```diff
@@ -49,7 +51,7 @@ struct BasicGateFuser final {
    * applied together. Note that gates fused with this method are not
    * multiplied together until ApplyFusedGate is called on the output.
    * @param num_qubits The number of qubits acted on by 'gates'.
-   * @param gates The gates to be fused.
+   * @param gates The gates to be fused. Gate times should be ordered.
```
95-martin-orion, line 54: This is a change from our previous assumptions, correct? re: #74

lib/fuser_basic.h
```diff
@@ -150,17 +188,66 @@ struct BasicGateFuser final {
...
+          while (last < times.size() && times[last] <= prev) ++last;
+        }
+      }
+
+      if (last == times.size()) break;
+    }
+
+    if (last < times.size()) {
+      times2.push_back(times[last]);
+    }
```
95-martin-orion, line 240: Is it possible that this could omit some of `times` from the returned value? Do we want this to happen?

lib/gate_appl.h
```diff
@@ -73,10 +73,10 @@ inline void ApplyGate(
     const Simulator& simulator, const Gate& gate, State& state) {
   typename Simulator::fp_type matrix[32];
 
-  if (gate.num_qubits == 1) {
+  if (gate.num_qubits == 1 && gate.matrix.size() == 8) {
     std::copy(gate.matrix.begin(), gate.matrix.begin() + 8, matrix);
     simulator.ApplyGate1(gate.qubits[0], matrix, state);
-  } else if (gate.num_qubits == 2) {
+  } else if (gate.num_qubits == 2 && gate.matrix.size() == 32) {
```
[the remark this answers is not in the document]
sergeisakov: Yes, neither `if` triggers for measurement gates. I don't really see any point to return an error here as measurement gates shouldn't be applied in these functions. I added two comments.

lib/gate_appl.h
```diff
@@ -96,16 +96,13 @@ inline void ApplyFusedGate(
     const Simulator& simulator, const Gate& gate, State& state) {
   typename Simulator::fp_type matrix[32];
 
-  if (gate.num_qubits == 1) {
+  if (gate.num_qubits == 1 && gate.pmaster->matrix.size() == 8) {
```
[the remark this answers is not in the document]
sergeisakov: `pmaster` is defined in `fuser.h`. It is a pointer to the main gate in a sequence of fused gates.

lib/statespace.h
```diff
@@ -94,9 +104,67 @@ class StateSpace {
...
+      norm += partial_norms[i];
+    }
+
+    return norm;
+  }
+
+  template <typename RGen, typename Func1, typename Func2>
+  MeasurementResult VirtualMeasure(
+      const std::vector<unsigned>& qubits, RGen& rgen, const State& state,
+      uint64_t size, Func1&& PartialNorms, Func2&& FindMeasuredBits) const {
```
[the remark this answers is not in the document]
sergeisakov: This is one of the ways of doing static polymorphism. I reimplemented this in a different way using the curiously recurring template pattern.

lib/statespace.h
```diff
@@ -94,9 +104,67 @@ class StateSpace {
...
+    for (std::size_t i = 1; i < partial_norms.size(); ++i) {
+      partial_norms[i] += partial_norms[i - 1];
+    }
+
+    auto norm = partial_norms.back();
+    auto r = RandomValue(rgen, norm);
+
+    unsigned m = 0;
+    while (r > partial_norms[m]) ++m;
+    r -= m > 0 ? partial_norms[m - 1] : 0;
```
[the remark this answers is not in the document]
sergeisakov: Done.

lib/parfor.h
```diff
@@ -24,6 +24,20 @@ namespace qsim {
 
 template <uint64_t MIN_SIZE>
 struct ParallelForT {
+  static bool CanBeParallel(uint64_t size) {
```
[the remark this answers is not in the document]
sergeisakov: Removed. I used this function in the first version of tests.

lib/parfor.h
```diff
@@ -24,6 +24,20 @@ namespace qsim {
...
+    return size >= MIN_SIZE;
+  }
+
+  static uint64_t GetIndex0(
```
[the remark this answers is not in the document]
sergeisakov: To get a measured bitstring, we need to sum up the bitsting probabilities until the sum does not exceed a given random number. We can reuse partial sums obtained at the previous step when we calculated the norm using threads. First, we find the first cumulative partial sum that exceeds our random number. This can be done quickly as the number of threads is much smaller than the number of bit strings. Then we iterate over bitstring indices beginning with the first index and ending with the last index that were used to obtain the required partial sum. That is we need to know exactly the same indices that were used by `ParallelFor` to divide work between threads.

lib/statespace_avx.h
sergeisakov: Done.

lib/statespace.h
sergeisakov: No, we don't use `For::Run` here. In principle, we can repeat the previous step a few times recursively before that but this is probably beyond this PR. There is a for loop in `FindMeasuredBits` that need to be run sequentially. It begins with `k0` and ends with `k1`. Those `k0` and `k1` should be the same indices that were used by thread number `m` in `For::RunReduce` at the previous step, see above.

lib/fuser_basic.h
sergeisakov: Done.

lib/fuser_basic.h
sergeisakov: Correct. Gates don't get explicitly sorted. So there was a possibility of getting incorrect results for unordered gate times.

lib/fuser_basic.h
sergeisakov: Yes, it is possible if all the remaining `times` are greater than `gates.back().time`. We don't need them.

sergeisakov:
Thank you for the review. I added a function comment.

PR #143, opened 2020-07-07
Convert to stateful for.

### LINKED ISSUE #142: [Proposal] Convert Templated XXXFor Structs to Objects
Right now both external and internal qsim simulator and statespace objects make use of structs with static templated methods in order to provide parallelization support. For example: https://github.com/quantumlib/qsim/blob/master/lib/parfor.h . In order for TensorFlow Quantum to finish transitioning over to this version of qsim from the one we had originally adapted we need to write our own struct that defines the static methods `Run` and `RunReduce` that make use of Tensorflow's threadpool.

Unfortunately unlike OpenMP, TF Threads are not global state that you can immediately use with some kind of `#pragma`. The way TensorFlow works is that when a C++ op is called it is supplied some portion of TensorFlow's available threads at call time. It is then our job to give these threads work to do. The problem is that because these threads are supplied to a local function at call time we can't make use of a global state and can't have our `Run` and `RunReduce` be truly static. I have tried some alternative like this:

```C++

struct ParallelForTFQ {
  static tensorflow::OpKernelContext* context = <pointer containing thread info resolved at call time>

  template <typename Function, typename... Args>
  static void Run(
      unsigned num_threads, uint64_t size, Function&& func, Args&&... args) {
        context->run(func (etc etc))
  }
```

Which had not let to much success because TFQ might have many instances of a qsim powered op running at once competing for and overwriting the `context` resource (digging into writing a resource lock compatible with TF threadpool isn't something I think we want to do either).

With all of this in mind I'm proposing that we move from templating the **type** of the XXXFOR struct we want a simulator and statespace to use, to instead supplying **instances** of the XXXFOR struct to the methods that require it. This would mean instead of doing something like:

```C++
For::Run(stuff)
```
We would do
```C++
<Template typename for_runner>
... my_function(const for_runner& for_obj, ...., ... ) {
.
.
.
for_obj.Run(stuff)
```

This seems like a simple way to allow one to incorporate the amount of local/object state needed for parallelization with TensorFlow. What do you @sergeisakov and @95-martin-orion think ? (This is a blocking issue for TFQ to transition over to this version of qsim)

sergeisakov:
Closes #142

### DIFF
--- a/lib/seqfor.h
+++ b/lib/seqfor.h
@@ -22,19 +22,18 @@
 namespace qsim {
 
 struct SequentialFor {
-  static uint64_t GetIndex0(
-      uint64_t size, unsigned num_threads, unsigned thread_id) {
+  explicit SequentialFor(unsigned num_threads) {}
+
+  static uint64_t GetIndex0(uint64_t size, unsigned thread_id) {
     return 0;
   }
 
-  static uint64_t GetIndex1(
-      uint64_t size, unsigned num_threads, unsigned thread_id) {
+  static uint64_t GetIndex1(uint64_t size, unsigned thread_id) {
     return size;
   }
 
   template <typename Function, typename... Args>
-  static void Run(
-    unsigned num_threads, uint64_t size, Function&& func, Args&&... args) {
+  static void Run(uint64_t size, Function&& func, Args&&... args) {
     for (uint64_t i = 0; i < size; ++i) {
       func(1, 0, i, args...);
     }
@@ -42,8 +41,7 @@ struct SequentialFor {
 
   template <typename Function, typename Op, typename... Args>
   static std::vector<typename Op::result_type> RunReduceP(
-      unsigned num_threads, uint64_t size, Function&& func, Op&& op,
-      Args&&... args) {
+      uint64_t size, Function&& func, Op&& op, Args&&... args) {
     typename Op::result_type result = 0;
 
     for (uint64_t i = 0; i < size; ++i) {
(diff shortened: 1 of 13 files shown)


lib/seqfor.h
```diff
@@ -22,28 +22,26 @@
 namespace qsim {
 
 struct SequentialFor {
-  static uint64_t GetIndex0(
-      uint64_t size, unsigned num_threads, unsigned thread_id) {
+  explicit SequentialFor(unsigned num_threads) {}
+
+  static uint64_t GetIndex0(uint64_t size, unsigned thread_id) {
```
95-martin-orion, line 27: Is there a benefit to leaving these methods as static? I could see it causing confusion if a future user was trying to create a "For" implementation and used this as reference.

lib/simulator_avx.h
```diff
@@ -32,8 +32,9 @@ class SimulatorAVX final {
   using State = typename StateSpace::State;
   using fp_type = typename StateSpace::fp_type;
 
-  SimulatorAVX(unsigned num_qubits, unsigned num_threads)
-      : num_qubits_(num_qubits), num_threads_(num_threads) {}
+  template <typename... Args>
+  explicit SimulatorAVX(unsigned num_qubits, Args&&... args)
```
95-martin-orion, line 36: Could we pass a pre-constructed For object to this constructor instead? That would give users the flexibility to configure their For object outside its constructor (e.g. if some parts of the For object could generate errors during setup).

If this makes tests overly verbose, a default value for the For argument could be used.

95-martin-orion: Likewise for other simulators and statespace objects.

lib/seqfor.h
sergeisakov: Well, the only benefit is that those functions can be called without creating an object. I guess there is also no benefit to make these methods non-static except for the reason you mentioned. I added a comment.

lib/simulator_avx.h
sergeisakov: One can pass a pre-constructed For object to this constructor. The following will work:
```C++
MyFor my_for(num_threads, ...);
SimulatorAVX<const MyFor&> simulator(num_qubits, my_for);
```
If `MyFor` has a copy constructor, the following will also work:
```C++
MyFor my_for(num_threads, ...);
SimulatorAVX<MyFor> simulator(num_qubits, my_for);

95-martin-orion:
@googlebot: rescan

PR #152, opened 2020-07-15
Add *SAN tests and config for tcmalloc.

### LINKED ISSUE #151: Support for tcmalloc.
Cloud colab for most of google uses tcmalloc instead of regular malloc. TFQ needs to be reliable under both. When running msan and asan tests for qsim, would it be possible to add a second pair of msan and asan tests where we run under `LD_PRELOAD ... libtcmalloc.so4 (or whatever it is)` ?

95-martin-orion:
Fixes #151. tcmalloc can be used by specifying `--config=tcmalloc` in the Bazel invocation.

Also expands the test matrix to include *SAN tests and tcmalloc/malloc variants. This might increase CI test times, but since Docker is the current bottleneck by a large margin I don't think that's a serious issue.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,7 +2,7 @@
 build --copt -std=c++11
 build --copt -D_GLIBCXX_USE_CXX11_ABI=0
 
-##### Sanitizers (can be added to any build) #####
+##### Sanitizers (choose one, or nosan for none) #####
 
 # Shared config for sanitizers
 build:sanitizer --strip=never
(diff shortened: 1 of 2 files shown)


95-martin-orion:
It doesn't look like the full test matrix was triggered for this PR, but I've locally verified that MSAN, ASAN, and tcmalloc variants of the tests each pass in isolation.

95-martin-orion:
> It doesn't look like the full test matrix was triggered for this PR, but I've locally verified that MSAN, ASAN, and tcmalloc variants of the tests each pass in isolation.

In retrospect, this is likely because the required tests no longer exist (the parameter list now has 4 elements, so e.g. "Test with Bazel (avx, basic)" cannot be found)

PR #157, opened 2020-07-22
Faster AVX simulator.

sergeisakov:
Make the AVX simulator faster for gates with qubit indices smaller than 3.


sergeisakov:
The performance is affected only for gates with qubit indices smaller than 3. Such gates can be faster by up to a factor of 2. The overall performance gain is of course much smaller. For instance, it is about 7% for `circuits/circuit_q30` (single thread).

PR #158, opened 2020-07-22
Faster SSE simulator.

sergeisakov:
Make the SSE simulator faster for gates with qubit indices smaller than 2.

### DIFF
--- a/lib/simulator_sse.h
+++ b/lib/simulator_sse.h
@@ -172,7 +172,8 @@ class SimulatorSSE final {
 
       auto p0 = rstate + 8 * i;
 
-      // 177 = 0b10110001, 78 = 0b01001110
+      // 177 = 0b10110001: shuffle four elements dcba -> cdab.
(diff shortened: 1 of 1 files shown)


lib/simulator_sse.h
```diff
@@ -138,88 +137,68 @@ class SimulatorSSE final {
...
-        break;
-      case 1:
-        rn = _mm_shuffle_ps(rn, rm, 68);  // 01000100
-        in = _mm_shuffle_ps(in, im, 68);
-        break;
-      default:
-        // Cannot reach here.
-        break;
-      }
+      // 177 = 0b10110001, 78 = 0b01001110
```
95-martin-orion, line 175: Could you add context to this comment or remove it?

sergeisakov: These numbers specify how to shuffle four elements in `__m128` registers. I added a short explanation.

PR #160, opened 2020-07-23
QSimSimulator now inherits from cirq.SimulatesSamples (and python support added for measurement gates)

TylerLeite:


TylerLeite:
> We found a Contributor License Agreement for you (the sender of this pull request), but were unable to find agreements for all the commit author(s) or Co-authors. If you authored these, maybe you used a different email address in the git commits than was used to sign the CLA ([login here](https://cla.developers.google.com/) to double check)? If these were authored by someone else, then they will need to sign a CLA as well, and confirm that they're okay with these being contributed to Google.
> In order to pass this check, please resolve this problem and then comment `@googlebot I fixed it.`. If the bot doesn't comment, it means it doesn't think anything has changed.
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fquantumlib%2Fqsim%2Fpull%2F160) for more info**.

@googlebot I fixed it!

qsimcirq/qsim_simulator.py
```diff
@@ -62,6 +64,124 @@ def __init__(self, qsim_options: dict = {}):
...
+    for i in range(repetitions):
+      qsim_state = qsim.qsim_simulate_fullstate(options)
+      amplitudes = QSimSimulatorState(qsim_state, qubit_map).state_vector
+
+      # Convert amplitudes to probabilities
+      probabilities = np.abs(amplitudes)**2
+      probabilities /= probabilities.sum()
+
+      # format the bitstring in terms of measurement gates
+      result = np.random.choice(bitstrings, p=probabilities)
```
95-martin-orion, line 177: Given that the state vector object has O(2^N) elements for N qubits, we want to avoid copying it or operating on it in Python. This will require a separate pybind method which extracts the results from a simulation.

Incidentally, measurements in qsim already calculate their results - we just don't store them. Preferably we would use these results in the pybind method.

TylerLeite:
> We found a Contributor License Agreement for you (the sender of this pull request), but were unable to find agreements for all the commit author(s) or Co-authors. If you authored these, maybe you used a different email address in the git commits than was used to sign the CLA ([login here](https://cla.developers.google.com/) to double check)? If these were authored by someone else, then they will need to sign a CLA as well, and confirm that they're okay with these being contributed to Google.
> In order to pass this check, please resolve this problem and then comment `@googlebot I fixed it.`. If the bot doesn't comment, it means it doesn't think anything has changed.
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fquantumlib%2Fqsim%2Fpull%2F160) for more info**.

@googlebot I fixed it.

95-martin-orion:
Discussed offline - until a proper C++ implementation for sampling is in place, the preferred method for this is to use `qsim_simulator.simulate(circuit)` and feed the resulting state vector into `cirq.sample_state_vector`: https://github.com/quantumlib/Cirq/blob/1b05c0ec860de8d61eacbb53e2968cd38ad5f815/cirq/sim/wave_function.py#L185

Keeping this PR open as it provides the Python "frontend" to the eventual C++ measurement support.

qsimcirq/qsim_simulator.py
95-martin-orion: PR #162 added the `qsim_sample` method, which returns a flat list of measurement values ordered by (time, qubit index). Calling this instead of `qsim_simulate_fullstate` should allow you to skip the measurement-extraction steps here (although converting to a dict will still be necessary).

Note that `qsim_sample` works for both terminal and non-terminal measurements, but must be called separately for each repetition since measurements are nondeterministic. For circuits with only terminal measurements, calling `simulate()` once and extracting results from the resulting state vector is likely faster.

PR #162, opened 2020-07-27
Expose measurement results for qsimcirq.

95-martin-orion:
This provides the C++ backend for #160. Some additional changes in that PR will be required to fully support measurement in the qsimcirq interface.

### DIFF
--- a/lib/run_qsim.h
+++ b/lib/run_qsim.h
@@ -123,17 +123,24 @@ struct QSimRunner final {
   }
 
   /**
-   * Runs the given circuit and make the final state available to the caller.
+   * Runs the given circuit and make the final state available to the caller,
+   * recording the result of any intermediate measurements in the circuit.
    * @param param Options for parallelism and logging.
    * @param circuit The circuit to be simulated.
    * @param state As an input parameter, this should contain the initial state
    *   of the system. After a successful run, it will be populated with the
    *   final state of the system.
+   * @param measure_results As an input parameter, this should be empty.
+   *   After a successful run, this will contain all measurements results from
+   *   the run, ordered by time and qubit index.
    * @return True if the simulation completed successfully; false otherwise.
    */
   template <typename Circuit>
-  static bool Run(const Parameter& param, const Circuit& circuit,
-                  typename Simulator::State& state) {
+  static bool Run(
+      const Parameter& param, const Circuit& circuit,
+      typename Simulator::State& state,
+      std::vector<typename Simulator::StateSpace::MeasurementResult>&
+          measure_results) {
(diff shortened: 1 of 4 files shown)


lib/run_qsim.h
```diff
@@ -123,17 +123,24 @@ struct QSimRunner final {
...
+   *   the run, ordered by time and qubit index.
    * @return True if the simulation completed successfully; false otherwise.
    */
   template <typename Circuit>
-  static bool Run(const Parameter& param, const Circuit& circuit,
-                  typename Simulator::State& state) {
+  static bool Run(
+      const Parameter& param, const Circuit& circuit,
+      typename Simulator::State& state,
+      std::vector<typename Simulator::StateSpace::MeasurementResult>&
```
sergeisakov, line 142: It might make sense to add `using MeasurementResult = typename Simulator::StateSpace::MeasurementResult;` above and replace `typename Simulator::StateSpace::MeasurementResult` with `MeasurementResult` for better formatting.

lib/run_qsim.h
```diff
@@ -123,17 +123,24 @@ struct QSimRunner final {
...
+          measure_results) {
```
sergeisakov, line 143: Perhaps it makes sense to add `measure_results.reserve(fused_gates.size());` below.

pybind_interface/pybind_main.cpp
```diff
@@ -351,6 +351,60 @@ py::array_t<float> qsim_simulate_fullstate(const py::dict &options) {
...
+  State state = state_space.CreateState(fsv);
+  state_space.SetStateZero(state);
+
+  if (!Runner::Run(param, circuit, state, results)) {
+    IO::errorf("qsim sampling of the circuit errored out.\n");
+    return {};
+  }
+
+  std::vector<bool> result_bits;
+  for (const auto& result : results) {
```
sergeisakov, line 401: Perhaps it makes sense to declare `bitstring` as `std::vector<bool>` instead of `std::vector<unsigned>` in lib/statespace.h to avoid this conversion.

lib/run_qsim.h
95-martin-orion: Done. Also updated the other `using` declarations affected by this.

lib/run_qsim.h
95-martin-orion: Done.

pybind_interface/pybind_main.cpp
95-martin-orion: Opted to convert `result_bits` to `std::vector<unsigned>` instead.

PR #167, opened 2020-08-06
Connect sampler to qsimcirq

95-martin-orion:
This PR includes the changes by @TylerLeite from #160 and connects them to the new sampling "backend" added in #162. Following this change, it will be possible to invoke circuit-sampling requests through qsimcirq, notably including circuits with intermediate (i.e. non-terminal) measurements.

Requesting multiple repetitions from the sampler will execute the circuit once for each sample, since intermediate measurements introduce nondeterminism in the final state.


pybind_interface/pybind_main.cpp
```diff
@@ -354,10 +363,8 @@ py::array_t<float> qsim_simulate_fullstate(const py::dict &options) {
 // TODO(95-martin-orion): add support for measurement gate conversion.
 std::vector<unsigned> qsim_sample(const py::dict &options) {
```
sergeisakov, line 364: It seems that measurements are not reproducible in Cirq. To emulate this behavior, a random seed should be set below.

qsimcirq/qsim_simulator.py
```diff
@@ -62,6 +64,120 @@ def __init__(self, qsim_options: dict = {}):
...
+
+  def _sample_bitstrings(
+    self,
+    program: circuits.Circuit,
+    repetitions: int = 1,
+  ) -> Dict[str, np.ndarray]:
+    """Samples bitstrings from the circuit
+
+    All MeasurementGates must be terminal.
+    Note that this does not collapse the wave function.
```
sergeisakov, line 97: We actually collapse the wave function. Is that consistent with what Cirq expects?

qsimcirq/qsim_simulator.py
```diff
@@ -62,6 +64,120 @@ def __init__(self, qsim_options: dict = {}):
...
+    results = {}
+    for key, bound in bounds.items():
+      boundlen = bound[1]-bound[0]
+      results[key] = np.ndarray(shape=(repetitions, bound[1]-bound[0]),
+                                dtype=int)
+
+    for i in range(repetitions):
+      measurements = qsim.qsim_sample(options)
+      print(f'measurements: {measurements}')
+      for key, bound in bounds.items():
```
sergeisakov, line 174: If the keys are iterated in the order of occurrence in the circuit then I guess all this might work correctly without any modifications here and in `pybind_main.cpp`.

pybind_interface/pybind_main.cpp
```diff
@@ -240,6 +240,15 @@ void add_gate(const qsim::Cirq::GateKind gate_kind, const unsigned time,
         Cirq::FSimGate<float>::Create(time, qubits[0], qubits[1],
                                       params.at("theta"), params.at("phi")));
       break;
+    case Cirq::kMeasurement: {
+      std::vector<unsigned> qubits_;
```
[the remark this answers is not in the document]

qsimcirq/qsim_simulator.py
```diff
@@ -62,6 +64,120 @@ def __init__(self, qsim_options: dict = {}):
...
+    Returns:
+        A dictionary from measurement gate key to measurement
+        results.
+    """
+    param_resolver = param_resolver or study.ParamResolver({})
+    solved_circuit = protocols.resolve_parameters(circuit, param_resolver)
+
+    return self._sample_bitstrings(solved_circuit, repetitions)
+
+  def _sample_bitstrings(
```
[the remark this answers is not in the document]
95-martin-orion: Renamed to `_sample_measurement_results` - does this give the right impression?

qsimcirq/qsim_simulator.py
```diff
@@ -62,6 +64,120 @@ def __init__(self, qsim_options: dict = {}):
...
+
+    Returns:
+        A dictionary from measurement gate key to measurement
+        results. Measurement results are stored in a 2-dimensional
+        numpy array, the first dimension corresponding to the repetition
+        and the second to the actual boolean measurement results (ordered
+        by the qubits being measured.)
+
+    Raises:
+        NotImplementedError: If there are non-terminal measurements in the
```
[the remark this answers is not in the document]
95-martin-orion: Removed.

qsimcirq/qsim_simulator.py
```diff
@@ -62,6 +64,120 @@ def __init__(self, qsim_options: dict = {}):
...
+    return self._sample_bitstrings(solved_circuit, repetitions)
+
+  def _sample_bitstrings(
+    self,
+    program: circuits.Circuit,
+    repetitions: int = 1,
+  ) -> Dict[str, np.ndarray]:
+    """Samples bitstrings from the circuit
+
+    All MeasurementGates must be terminal.
```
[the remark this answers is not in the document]
95-martin-orion: Removed.

pybind_interface/pybind_main.cpp
95-martin-orion: Added seed pass-through behavior for both this case and the full state-vector case.

Since the random value generation for Cirq and qsim are in Python and C++ respectively, I don't think there's a clean way to get identical results from each even if the same seed is used. However, qsim results should be reproducible in qsim now.

qsimcirq/qsim_simulator.py
95-martin-orion: Removed this comment. Cirq does expect intermediate measurements to collapse the wave function.

qsimcirq/qsim_simulator.py
95-martin-orion: I'm not sure we can rely on this being the case. Cirq is pretty permissive with how users can insert measurements into the circuit, so I'd feel more comfortable keeping this conversion explicit.

qsimcirq/qsim_simulator.py
```diff
@@ -51,16 +53,124 @@ def __init__(self,
...
 
-class QSimSimulator(SimulatesAmplitudes, SimulatesFinalState):
+class QSimSimulator(SimulatesSamples, SimulatesAmplitudes, SimulatesFinalState):
 
-  def __init__(self, qsim_options: dict = {}):
+  def __init__(self, qsim_options: dict = {},
+               seed: value.RANDOM_STATE_OR_SEED_LIKE = None):
     if any(k in qsim_options for k in ('c', 'i')):
       raise ValueError(
           'Keys "c" & "i" are reserved for internal use and cannot be used in QSimCircuit instantiation.'
```
sergeisakov, line 62: Should "t" and "v" be added here?

pybind_interface/pybind_main.cpp
sergeisakov: Sure, there is no way to get identical results. I just meant that it seems that Cirq gives different results from invocation to invocation. qsim results depend on the provided seed and previously the results were always the same because the default seed was used.

qsimcirq/qsim_simulator.py
sergeisakov: Disregard my comment. It should be OK even if the keys are not ordered.

qsimcirq/qsim_simulator.py
```diff
@@ -51,16 +53,124 @@ def __init__(self,
...
+    indices = [qubit_map[qubit] for qubit in measured_qubits]
+
+    # Set qsim options
+    options = {}
+    options.update(self.qsim_options)
+    options['c'] = program.translate_cirq_to_qsim(ops.QubitOrder.DEFAULT)
+
+    results = {}
+    for key, bound in bounds.items():
+      boundlen = bound[1]-bound[0]
```
sergeisakov, line 161: `boundlen` is not used.

qsimcirq/qsim_simulator.py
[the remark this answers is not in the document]
sergeisakov: Yes.

qsimcirq/qsim_simulator.py
95-martin-orion: Removed.

qsimcirq/qsim_simulator.py
95-martin-orion: Done.

95-martin-orion: Reverted this change - the "t" and "v" values here are only default values. Users can set them manually.

95-martin-orion:
I have manually verified that @TylerLeite signed the CLA, but the automatic verifier isn't picking it up. Added `cla: yes` label to override and submit.

Issue #129, opened 2020-06-09
Qsim Cirq tutorial


karlunho:
Need to put together a colab showing how to use Cirq and qsim together.

95-martin-orion:
@alexandrupaler has created something similar for Cirq simulator workshops; we might consider promoting one of those to be the "canonical" tutorial.

Issue #118, opened 2020-05-29
qsimcirq needs better documenation support for "compute_amplitudes"


karlunho:
The function "compute_amplitudes" does not exist in the Cirq simulator interface, so it doesn't really make sense for the qsimcirq API to support this function. We should either add "compute_amplitudes" to the cirq simulator interface or remove this function for now.

95-martin-orion:
The `SimulatesAmplitudes` interface has this method:
https://github.com/quantumlib/Cirq/blob/eb4ad794a9213c20c800fa59199a29f353ca7994/cirq/sim/simulator.py#L126

That said, it does recommend implementing "compute_amplitudes_sweep" instead - we may want to shift to that method.

karlunho:
Qsimcirq simulator should API documentation on what is supported.

95-martin-orion:
This documentation will be available in the upcoming devsite.

95-martin-orion:
Duplicate of #120

PR #168, opened 2020-08-10
Update usage and qsimcirq docs.

95-martin-orion:
This brings the qsimcirq and bazel docs up to speed and adds a bit of clarification to the usage docs.

@karlunho: I'm aware of e.g. #129 (tutorial colab) and #118 (API docs) - is there a plan for these already? Are there any other docs concerns that need to be addressed in this PR?

### DIFF
--- a/docs/cirq_interface.md
+++ b/docs/cirq_interface.md
@@ -3,7 +3,7 @@
 This file provides examples of how to use qsim with the
 [Cirq](https://github.com/quantumlib/cirq) Python library.
 
-qsim is currently built to work with Cirq version 0.6.0; if you have a later
+qsim is currently built to work with Cirq version 0.8.2; if you have a later
 version of Cirq installed, you may need to downgrade (or run in a virtualenv)
 when working with qsim until support for the latest Cirq version is available.
 
@@ -43,9 +43,8 @@ directory.
 ```
 make run-py-tests
 ```
-This will run [qsim_test](/qsimcirq_tests/qsim_test.py), which calls the Python
-binary directly, and [qsimcirq_test](/qsimcirq_tests/qsimcirq_test.py), which
-invokes qsim through the qsim-Cirq interface.
+This will run [qsimcirq_test](/qsimcirq_tests/qsimcirq_test.py), which invokes
+qsim through the qsim-Cirq interface.
 
 ## Interface design and operations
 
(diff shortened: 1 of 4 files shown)


docs/usage.md
```diff
@@ -160,7 +178,30 @@ format.
 
 Example:
 ```
+./qsimh_amplitudes.x -c ../circuits/circuit_q40 -d 47 -k 0,1,2,3,4,5,6,7,8,9,10,13,14,15,16,17,23,24 -t 8 -w 0 -p 0 -r 4 -i ../circuits/bitstrings_q40_s1 -o ampl_q40_s1_w0 -v 1
```
sergeisakov, line 181: Please replace `-r 4` with `-r 13` and `ampl_q40_s1_w0` with `ampl_q40_s1`. It makes sense to describe how `-r` affects performance (perhaps in a later PR). There are 15 gates on the cut and there are two gates on the cut in the last cycle of two-qubit gates in this particular case. For better performance, it makes sense to perform the second checkpoint before these two gates.
In addition, could you perhaps replace `Usage` with `Usage of sample applications` or something like that in line 1?

docs/usage.md
```diff
@@ -160,7 +178,30 @@ format.
...
+```
+Note that this could take a very long time to run, since parallelism on a
```
sergeisakov, line 183: It will take a very long time (a few weeks or so). This is a large circuit with 15 fSim gates on the cut. There are 4^15 paths in total.

docs/usage.md
95-martin-orion: Done.

docs/cirq_interface.md
```diff
@@ -79,14 +74,14 @@ QSimSimulator uses a Schrödinger full state-vector simulator, suitable for
 acquiring the complete state of a reasonably-sized circuit (~35 qubits):
 ```
 my_sim = qsimcirq.QSimSimulator()
```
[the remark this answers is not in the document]
95-martin-orion: Done.

docs/usage.md
95-martin-orion: Made the comment more explicit, and added warnings against running this command directly.

docs/usage.md
```diff
@@ -111,20 +109,60 @@ Example:
...
+Deciding which cuts are optimal for a given circuit is computationally hard.
+However, splitting the grid into roughly equal parts with the fewest cuts
+possible (as is done for the lattice above) produces a circuit that performs
+reasonably well in most cases.
+
+**-p** and **-r** specify the number of "prefix" and "root" gates respectively,
+and determine how much of the circuit must be repeated for each execution.
+Values for these flags should be chosen with this in mind:
+
+- Operations up to and including the "prefix" gates will only be run once, but
```
sergeisakov, line 158: These lines might be a bit confusing.

95-martin-orion: Removed a line - the list below explains it better. Let me know if this didn't fix your concern.

sergeisakov: Perhaps I would add something like the following:

The runtime can be controlled by **-p**. There is no summation over the "prefix" gates and the unique "prefix" path is specified by **-w**, see the `Distributed execution` section below for details on this. **-r** specifies implicitly the number of the "suffix" gates: the total number of gates on the cut minus the values specified by **-p** and **-r**. Typically for performance, the "suffix" gates should be the gates on the cut with maximum "time".

95-martin-orion: Done.

PR #173, opened 2020-08-13
Add auxiliary statespace functions.

sergeisakov:


sergeisakov:
> Implementation looks good at a glance - could you add a short test to verify these behave as expected?

Done.

PR #178, opened 2020-08-17
State add

MichaelBroughton:
Added the ability to add two state vector together. This would be another useful downstream feature for TFQ.

### DIFF
--- a/lib/statespace_avx.h
+++ b/lib/statespace_avx.h
@@ -229,7 +229,7 @@ struct StateSpaceAVX : public StateSpace<StateSpaceAVX<For>, For, float> {
   }
 
   // Does the equivalent of a += b elementwise.
-  void AddState(const state& a, const state& b) {
+  void AddState(const State& a, const State& b) {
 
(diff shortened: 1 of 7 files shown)


lib/statespace_avx.h
```diff
@@ -228,6 +228,23 @@ struct StateSpaceAVX : public StateSpace<StateSpaceAVX<For>, For, float> {
     state.get()[p + 8] = im;
   }
 
+  // Does the equivalent of a += b elementwise.
+  void AddState(const State& a, const State& b) {
```
sergeisakov, line 232: Could you switch a and b here and in lib/statespace_basic.h and lib/statespace_sse.h to conform to the convention that input-only parameters are placed before output parameters.

95-martin-orion: If we do this, could the arguments also be given more descriptive names (e.g. source, dest)?

MichaelBroughton: Changed names and order.

PR #182, opened 2020-08-19
Added const to AddState functions.

MichaelBroughton:
Forgot these when first putting in `AddState`. Any chance we could cut a 0.3.1 to add these in :) ? ( I got hit by this right in the middle of a big TFQ feature ).


95-martin-orion:
> Forgot these when first putting in `AddState`. Any chance we could cut a 0.3.1 to add these in :) ? ( I got hit by this right in the middle of a big TFQ feature ).

Sure - would you mind also updating the version to `0.3.1` in `setup.py` in this PR? This is a bugfix, so a patch version is fine.

PR #186, opened 2020-09-02
Update authors.

sergeisakov:


sergeisakov:
> I'm curious - do you plan to add a more comprehensive AUTHORS file (like the [Cirq AUTHORS file](https://github.com/quantumlib/Cirq/blob/master/AUTHORS)), or is that unnecessary here?

I am not sure. We can add an AUTHORS file.

balopat:
FWIW: AUTHORS is a little bit different - it is used for copyright note - so that we can write in the copyright "Copyright The Cirq Developers" - https://opensource.google/docs/releasing/authors/ - Maintaining a separate contributor list is questionable - everyone can see the contributors on Github. I guess that's why @sergeisakov removed the authors, right?

sergeisakov:
Yes, right.

PR #187, opened 2020-09-02
First pass on api docs.

MichaelBroughton:
Adds in the api docs build scripts. I had a lot of trouble running this, for some reason qsim refused to work on my corp laptop (Kept getting circular import issues after build).

@95-martin-orion would you be willing to checkout this code and do:

```bash
python3 -m pip install git+https://github.com/tensorflow/docs
python3 build_docs.py
ls /tmp/qsim_api
```
and see if stuff shows up there ?

If it does that means it works and we can start to iterate on how things look.


95-martin-orion:
> Adds in the api docs build scripts. I had a lot of trouble running this, for some reason qsim refused to work on my corp laptop (Kept getting circular import issues after build).

I'm seeing a similar issue. This sounds like the "import file vs. import installed package" problem - there's a note about this in the [install doc](https://github.com/quantumlib/qsim/blob/master/docs/install_qsimcirq.md#testing). Essentially, you'll need to move `build_api_docs.py` outside of the qsim directory in order to run it.

If that's not an option, we may need to revisit the import hierarchy in qsimcirq (similar to the [Cirq init file](https://github.com/quantumlib/Cirq/blob/master/cirq/__init__.py)).

MichaelBroughton:
Oops, thanks for the tip. Should be good to go now.

95-martin-orion (changes requested):
That seems to have fixed the original issue, but now I'm seeing:
```
{ ...stack trace... }
File "/usr/local/google/home/martinop/venv-qsim/lib/python3.8/site-packages/tensorflow_docs/api_generator/public_api.py", line 249, in _is_private
    if name in self._private_map.get('.'.join(path), []):
AttributeError: 'set' object has no attribute 'get'
```

What does your setup process look like? I've been running this in a virtualenv, only installing qsimcirq and the tensorflow_docs as indicated.

lamberta:
cc: @markdaoust

build_api_docs.py
```diff
@@ -0,0 +1,62 @@
...
+
+    doc_generator = generate_lib.DocGenerator(
+        root_title="qsim",
+        py_modules=[("qsimcirq", qs)],
+        base_dir=os.path.dirname(qs.__file__),
+        code_url_prefix=FLAGS.code_url_prefix,
+        search_hints=FLAGS.search_hints,
+        site_path=FLAGS.site_path,
+        callbacks=[public_api.local_definitions_filter],
+        private_map={"""Block things from showing up here."""})
```
MarkDaoust: The problem is with this line. This `private_map` argument expects a dict. This is a set. What template did you get this from? we should make this clearer.

If you're not  using this argument, you don't need to pass it. You can just delete this line.

```suggestion
        )
```

MichaelBroughton: Switched over to my work machine. Got the same error, with and without the private_map argument. Can anyone else confirm this ?

MarkDaoust: I just tried this locally by cloning from your fork and `pip install  .`, 

As output I got:

```
$ python docs/build_api_docs.py
$ tree /tmp/qsim_api
/tmp/qsim_api
├── index.md
├── qsimcirq
│   ├── _api_cache.json
│   ├── qsim
│   │   ├── add_gate.md
│   │   ├── add_matrix1.md
│   │   ├── add_matrix2.md
│   │   ├── Circuit.md
│   │   ├── GateKind.md
│   │   ├── qsimh_simulate.md
│   │   ├── qsim_sample.md
│   │   ├── qsim_simulate_fullstate.md
│   │   └── qsim_simulate.md
│   ├── qsim_circuit
│   │   └── QSimCircuit.md
│   ├── qsim_circuit.md
│   ├── qsimh_simulator
│   │   └── QSimhSimulator.md
│   ├── qsimh_simulator.md
│   ├── qsim.md
│   ├── qsim_simulator
│   │   ├── QSimSimulator.md
│   │   ├── QSimSimulatorState.md
│   │   └── QSimSimulatorTrialResult.md
│   └── qsim_simulator.md
├── qsimcirq.md
├── _redirects.yaml
└── _toc.yaml
```

MichaelBroughton: Interesting.... @95-martin-orion what do you want to do here ?

95-martin-orion: After the changes from @MarkDaoust, I now understand the process to be:
```
# from the top-level qsim directory
virtualenv ~/venv
source ~/venv/bin/activate
pip3 install .
python3 -m pip install git+https://github.com/tensorflow/docs
python3 docs/build_docs.py
tree /tmp/qsim_api
```
I'll try to reproduce this. Seeing different results between users is worrying, but in the end we really only need to worry about whether it works on the devsite.

95-martin-orion: Using the process above, I was able to produce the intended output.

95-martin-orion (approved):

PR #188, opened 2020-09-03
Add initial support for multi-qubit gates.

sergeisakov:
This PR also moves applying measurement gates from run_qsim.h to gate_appl.h.

### DIFF
--- a/lib/gate_appl.h
+++ b/lib/gate_appl.h
@@ -18,6 +18,7 @@
 #include <utility>
 #include <vector>
 
+#include "gate.h"
 #include "matrix.h"
 
 namespace qsim {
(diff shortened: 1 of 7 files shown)


lib/gate_appl.h
```diff
@@ -62,50 +63,150 @@ inline void CalcMatrix4(unsigned q0, unsigned q1,
...
+    }
+  } else {
+    ApplyGate(simulator, gate, state);
   }
+
+  return true;
+}
+
+/**
+ * Applies the given gate to the simulator state.
```
95-martin-orion, line 115: This comment should explicitly note that the function discards measurement results.

lib/gate_appl.h
```diff
@@ -62,50 +63,150 @@ inline void CalcMatrix4(unsigned q0, unsigned q1,
...
+    }
+  } else {
+    ApplyFusedGate(simulator, gate, state);
+  }
+
+  return true;
+}
+
+/**
+ * Applies the given fused gate to the simulator state.
```
95-martin-orion, line 193: This comment should explicitly note that the function discards measurement results.

95-martin-orion (approved):

lib/gate_appl.h
sergeisakov: Done.

lib/gate_appl.h
sergeisakov: Done.

PR #184, opened 2020-08-27
Update toc/book and tutorial

lamberta:
stage: cl/328835209

PR #189, opened 2020-09-08
Docstrings for gate types.

95-martin-orion:
This builds upon #184 by adding descriptions for each gate in the docs site. Some of these are admittedly self-explanatory, but I opted to provide a description for every gate for visual consistency.

### DIFF
--- a/lib/gates_cirq.h
+++ b/lib/gates_cirq.h
@@ -83,6 +83,9 @@ constexpr double is2_double = 0.7071067811865475;
 
 // Gates from cirq/ops/identity.py:
 
+/**
+ * A one-qubit identity gate.
+ */
 template <typename fp_type>
 struct I {
   static constexpr GateKind kind = kI;
@@ -95,6 +98,9 @@ struct I {
   }
 };
 
+/**
+ * A two-qubit identity gate.
+ */
 template <typename fp_type>
 struct I2 {
   static constexpr GateKind kind = kI2;
@@ -118,6 +124,9 @@ struct I2 {
 
 // Gates form cirq/ops/common_gates.py:
 
+/**
+ * A gate that rotates around the X axis of the Bloch sphere.
+ */
 template <typename fp_type>
 struct XPowGate {
   static constexpr GateKind kind = kXPowGate;
@@ -139,6 +148,9 @@ struct XPowGate {
   }
 };
 
+/**
+ * A gate that rotates around the Y axis of the Bloch sphere.
+ */
 template <typename fp_type>
 struct YPowGate {
   static constexpr GateKind kind = kYPowGate;
@@ -160,6 +172,9 @@ struct YPowGate {
   }
 };
 
+/**
+ * A gate that rotates around the Z axis of the Bloch sphere.
+ */
 template <typename fp_type>
 struct ZPowGate {
   static constexpr GateKind kind = kZPowGate;
@@ -181,6 +196,9 @@ struct ZPowGate {
   }
 };
 
+/**
+ * A gate that rotates around the X+Z axis of the Bloch sphere.
+ */
 template <typename fp_type>
 struct HPowGate {
   static constexpr GateKind kind = kHPowGate;
@@ -206,6 +224,9 @@ struct HPowGate {
   }
 };
 
+/**
+ * A gate that applies a phase to the |11⟩ state of two qubits.
+ */
 template <typename fp_type>
 struct CZPowGate {
   static constexpr GateKind kind = kCZPowGate;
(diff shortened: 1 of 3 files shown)


lib/gates_cirq.h
```diff
@@ -285,8 +309,10 @@ struct CXPowGate {
   }
 };
 
-// The (exponent=phi/pi, global_shift=-0.5) instance of XPowGate.
-// This is a function in Cirq.
+/**
```
sergeisakov, line 312: It might be worth adding some additional info for many gates (H, S, T, CZ, CZ, X, Y, Z, XX, YY, ZZ, SWAP, ISWAP, rx, ry, rz, and perhaps PhasedXPowGate, PhasedXZGate, PhasedISwapPowGate, riswap, givens) as in the Cirq API reference. For instance, "The Hadamard gate." can be added to the description of the H gate.

lib/gates_cirq.h
```diff
@@ -1073,6 +1160,9 @@ struct FSimGate {
 
 // Gates from cirq/ops/matrix_gates.py:
 
+/**
+ * A unitary one-qubit gate defined entirely by its matrix.
```
sergeisakov, line 1164: Strictly  speaking, MatrixGate1 and MatrixGate2 gates do not have to be unitary in general.

sergeisakov (approved):

lib/gates_cirq.h
95-martin-orion: I will fill out the additional detail here and merge this PR prior to cutting the new version. This should be ready later today.

lib/gates_cirq.h
95-martin-orion: Done.

lib/gates_cirq.h
95-martin-orion: I've added some detail to each of these, but tried to keep them concise (e.g. no matrix definitions for gates).

lib/gates_cirq.h
```diff
@@ -285,8 +315,11 @@ struct CXPowGate {
   }
 };
 
-// The (exponent=phi/pi, global_shift=-0.5) instance of XPowGate.
-// This is a function in Cirq.
+/**
+ * The `(exponent = phi/pi, global_shift = -0.5)` instance of XPowGate.
+ * This is equivalent to an X gate with a global phase.
```
sergeisakov, line 320: I think this is not equivalent to an X gate with a global phase, similarly for ry and rz. It is basically XPowGate with a fixed global shift. The  matrix is e^{-i X rads / 2}. In a sense, this is also a generalization of the X gate. This is the same as GateRX in gates_qsim.h.

lib/gates_qsim.h
```diff
@@ -230,6 +269,10 @@ struct GateRXY {
   }
 };
 
+/**
+ * A pi / 2 rotation around the X + Y axis.
+ * Analogous to the Hadamard gate, which rotates around the X + Z axis.
```
sergeisakov, line 274: I am not sure if this is analogous to the Hadamard gate, which rotates around the X + Z axis. This gate is a pi / 2 rotation around the X + Y axis with a global phase. The matrix is e^{i \pi / 4} e^{-i (X + Y) / \sqrt{2} \pi / 4}.

lib/gates_cirq.h
```diff
@@ -828,8 +924,11 @@ struct ISwapPowGate {
   }
 };
 
-// The (exponent=2*phi/pi, global_shift=0) instance of ISwapPowGate.
-// This is a function in Cirq.
+/**
+ * The `(exponent = 2*phi/pi, global_shift = 0)` instance of ISwapPowGate.
+ * This is equivalent to an ISWAP gate with a global phase.
```
sergeisakov, line 929: I think this is not equivalent to an ISWAP gate with a global phase. It is basically ISwapPowGate without a global shift.

lib/gates_cirq.h
```diff
@@ -670,7 +750,10 @@ struct ZZPowGate {
   }
 };
 
-// The (exponent=1, global_shift=0) instance of XXPowGate.
+/**
+ * The `(exponent = 1, global_shift = 0)` instance of XXPowGate.
+ * This is the X-parity gate.
```
sergeisakov, line 755: One can write "The tensor product of two X gates" instead of "This is the X-parity gate", similarly for YY and ZZ.

sergeisakov (approved):

lib/gates_cirq.h
95-martin-orion: Done (also ry and rz).

lib/gates_qsim.h
95-martin-orion: Removed the reference to Hadamard gate.

lib/gates_cirq.h
95-martin-orion: Done.

lib/gates_cirq.h
95-martin-orion: Done (also updated XX/YY/ZZPowGate).

sergeisakov (approved):
Looks good to me.

PR #190, opened 2020-09-09
Convert to QSimCircuit after resolving params.

95-martin-orion:
This also resolves a long-standing question of whether qsim supports parameterized circuits: now it does!

### DIFF
--- a/qsimcirq/qsim_circuit.py
+++ b/qsimcirq/qsim_circuit.py
@@ -126,7 +126,6 @@ class QSimCircuit(cirq.Circuit):
     return super().__eq__(other)
 
   def _resolve_parameters_(self, param_resolver: cirq.study.ParamResolver):
-
     return QSimCircuit(
       super()._resolve_parameters_(param_resolver), device=self.device)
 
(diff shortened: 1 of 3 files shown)


95-martin-orion:
TODO(@95-martin-orion): cut a new version and release qsimcirq after this PR. (May wait for a couple other changes as well)

sergeisakov (approved):
Should `_resolve_parameters_` be removed from qsim_circuit.py? It seems it is not used now and potentially it could lead to bugs.

95-martin-orion:
> Should `_resolve_parameters_` be removed from qsim_circuit.py? It seems it is not used now and potentially it could lead to bugs.

This is a good point - we really should handle the `_resolve_parameters_` return type in the qsim implementation. Restructured the PR to handle it that way instead - please take a look.

qsimcirq/qsim_circuit.py
```diff
@@ -127,11 +127,8 @@ def __eq__(self, other):
 
   def _resolve_parameters_(self, param_resolver: cirq.study.ParamResolver):
 
-    qsim_circuit = super()._resolve_parameters_(param_resolver)
-
-    qsim_circuit.device = self.device
-
-    return qsim_circuit
+    return QSimCircuit(
```
sergeisakov, line 130: Looks good to me. Perhaps empty line 129 should be deleted.

sergeisakov (approved):

qsimcirq/qsim_circuit.py
95-martin-orion: Done.

PR #192, opened 2020-09-10
Optimize all-terminal case for 'run'.

### LINKED ISSUE #191: QSimSimulator.run should only simulate once when there are no intermediate measurements
Instead, it prints out
```
Provided circuit has no intermediate measurements. It may be faster to sample from the final state vector. Continuing with one-by-one sampling.
```
Is there a reason this simple optimization was not implemented?

95-martin-orion:
Fixes #191. cc: @kevinsung 

The existing tests have good coverage for intermediate vs. terminal measurements, so I haven't added any new ones in this PR.

### DIFF
--- a/qsimcirq/qsim_simulator.py
+++ b/qsimcirq/qsim_simulator.py
@@ -118,14 +118,8 @@ class QSimSimulator(SimulatesSamples, SimulatesAmplitudes, SimulatesFinalState):
     if not isinstance(program, qsimc.QSimCircuit):
       program = qsimc.QSimCircuit(program, device=program.device)
 
-    if program.are_all_measurements_terminal() and repetitions > 1:
-      print('Provided circuit has no intermediate measurements. ' +
-            'It may be faster to sample from the final state vector. ' +
-            'Continuing with one-by-one sampling.')
-
     # Compute indices of measured qubits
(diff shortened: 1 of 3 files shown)


qsimcirq/qsim_simulator.py
```diff
@@ -118,14 +118,8 @@ def _sample_measure_results(
...
       program = qsimc.QSimCircuit(program, device=program.device)
 
-    if program.are_all_measurements_terminal() and repetitions > 1:
-      print('Provided circuit has no intermediate measurements. ' +
-            'It may be faster to sample from the final state vector. ' +
-            'Continuing with one-by-one sampling.')
-
     # Compute indices of measured qubits
     ordered_qubits = ops.QubitOrder.DEFAULT.order_for(program.all_qubits())
-    ordered_qubits = list(reversed(ordered_qubits))
```
sergeisakov, line 128: Why this line is not required?

sergeisakov (approved):

qsimcirq/qsim_simulator.py
95-martin-orion: The state vector returned by `qsim_simulate_fullstate` is also in qsim (reverse) ordering, so the two cancel out. `_sample_measure_results`, has similar behavior.

(Note that `ordered_qubits` and `qubit_map` are only used by the all-terminal-measurements case)

PR #194, opened 2020-09-16
Update _book.yaml to point to /reference/cc/qsim

yashk2810:


yashk2810:
@95-martin-orion @lamberta

95-martin-orion:
What is the interaction between this and #187? Is there additional work required to enable the Python docs?

yashk2810:
For Python API docs, there is a bug that's waiting on your reply. After that we can enable the python API docs.

95-martin-orion:
> For Python API docs, there is a bug that's waiting on your reply. After that we can enable the python API docs.

Commented there - the PR mentioned in the bug was #187, which has been merged.

yashk2810:
Okay -- @lamberta will setup the python stuff, but this change is orthogonal to python api docs.

95-martin-orion (approved):
I guess this is a temporary patch until the Python docs are fully enabled?

yashk2810:
The commented out part is a patch.

THe _doxygen.yaml yaml is not. That's permanent.

yashk2810:
@lamberta is going to fix this with the Python and C++ changes. Closing this PR.

PR #197, opened 2020-09-21
Make gate_appl dependency explicit

95-martin-orion:
Required for internal Bazel build process. I'm surprised this worked on external Bazel, actually - I don't see how gate_appl.h was able to resolve its dependency on gate.h when its only deps item was `:matrix`, which has no dependency on gate.h...

### DIFF
--- a/lib/BUILD
+++ b/lib/BUILD
@@ -175,7 +175,10 @@ cc_library(
 
 cc_library(
     name = "gate_appl",
-    hdrs = ["gate_appl.h"],
+    hdrs = [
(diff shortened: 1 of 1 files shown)


sergeisakov (approved):
Will it work if you add ":gate" to deps instead of adding "gate.h" to hdrs?

95-martin-orion:
> Will it work if you add ":gate" to deps instead of adding "gate.h" to hdrs?

Yes, and that's a more accurate representation of what's being done. Change applied.

PR #198, opened 2020-09-22
Adds remaining API docs

95-martin-orion:
Fills in the rest of the public C++ API for the qsim devsite page.

### DIFF
--- a/lib/circuit_qsim_parser.h
+++ b/lib/circuit_qsim_parser.h
@@ -27,8 +27,10 @@ namespace qsim {
 
 /**
  * Parser for the (deprecated) qsim <a href="https://github.com/quantumlib/qsim/blob/master/docs/input_format.md">file input format</a>.
- * New code should prefer <a href="https://github.com/quantumlib/Cirq">Cirq</a>
- * for designing circuits to simulate with qsim.
+ * The primary supported interface for designing circuits to simulate with qsim
+ * is <a href="https://github.com/quantumlib/Cirq">Cirq</a>, which relies on
+ * the Python-based qsimcirq interface. For C++ applications, Cirq gates can be
+ * explicitly constructed in code.
  */
 template <typename IO>
 class CircuitQsimParser final {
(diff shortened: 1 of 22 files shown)


lib/hybrid.h
```diff
@@ -68,19 +70,57 @@ struct HybridSimulator final {
...
+     */
+    unsigned num_gatexs;
   };
 
+  /**
+   * User-specified parameters for hybrid simulation.
+   */
   struct Parameter {
+    /**
+     * Fixed bitstring indicating values to assign to indices of prefix gates.
```
sergeisakov, line 112: It might be confusing in this form. `prefix` contains indices (one or two bits per index) of one-qubit gates in the Schmidt decompositions of prefix gates, i.e. `prefix` specifies the path for prefix gates.

sergeisakov (changes requested):

lib/circuit_qsim_parser.h
```diff
@@ -25,6 +25,11 @@
 
 namespace qsim {
 
+/**
+ * Parser for the (deprecated) qsim <a href="https://github.com/quantumlib/qsim/blob/master/docs/input_format.md">file input format</a>.
+ * New code should prefer <a href="https://github.com/quantumlib/Cirq">Cirq</a>
```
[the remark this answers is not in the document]
95-martin-orion: My intent here is that Cirq is the primary supported interface for creating circuits. I'll add a note indicating that this is Python-specific, and indicate how to set up circuits for C++ code.

lib/hybrid.h
95-martin-orion: My understanding is that each prefix gate on the cut gets decomposed into a pair of one-qubit gates with one or two "extra" shared indices, and that `prefix` contains the values assigned to those indices. Is this incorrect? Would labeling the indices as "cut indices of prefix gates" be clearer?

sergeisakov: Yes, each prefix gate on the cut gets decomposed as a sum of pairs of one-qubit gates. The index we are talking about is a term index in this sum. So perhaps "Schmidt decomposition indices of prefix gates" will be clearer.

sergeisakov (approved):

PR #199, opened 2020-09-25
Enforce gate order in qsimcirq.

95-martin-orion:
Previously, decomposed gates were added to the circuit out-of-order. This caused core qsim to ignore these gates, which can be seen by running the updated test from this PR without the other changes.

Time-ordering is now enforced when decomposing gates.


balopat (approved):
LGTM

PR #203, opened 2020-10-05
Add overview doc

95-martin-orion:
Requested for the upcoming devsite. Compare with [TFQ overview doc](https://www.tensorflow.org/quantum/overview).

\+ @karlunho for comments.

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -1,0 +1,56 @@
+# qsim and qsimh
+
+qsim and qsimh are a collection of C++ libraries for quantum circuit
+simulation. These libraries provide powerful, low-cost tools for
+researchers to test quantum algorithms before running on quantum hardware.
+
+qsim makes use of AVX/FMA vector operations, OpenMP multithreading, and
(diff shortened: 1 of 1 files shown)


docs/overview.md
```diff
@@ -0,0 +1,56 @@
...
+[install guide](https://github.com/quantumlib/qsim/blob/master/docs/install_qsimcirq.md)
+or try the runnable
+[notebook tutorial](https://github.com/quantumlib/qsim/blob/master/docs/tutorials/qsimcirq.ipynb).
+
+## Design
+
+This repository includes two top-level libraries for simulation:
+
+-   **qsim** is a Schrödinger state-vector simulator designed to run on a
+    single machine. It produces the full state vector as output, allowing
```
sergeisakov, line 25: I would probably add "... output, for instance, allowing ..." or something like that as the full state vector allows to do other things like calculating observables. You don't really have to add that. What you wrote is also good.

sergeisakov (approved):

PR #204, opened 2020-10-05
Adds zenodo config

balopat:
Adds zenodo config

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,3 +108,21 @@ More detailed information about the qsim-Cirq API can be found in the
 ## Disclaimer
 
 This is not an officially supported Google product.
+
+# How to cite qsim
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -108,3 +108,21 @@ More detailed information about the qsim-Cirq API can be found in the
 ## Disclaimer
 
 This is not an officially supported Google product.
+
+# How to cite qsim
```
95-martin-orion, line 112: I don't think the README page appears on the devsite. If we want this section to show up, it may need to go in a different doc.

balopat: That is fine. All the projects will have and should have a section about citation on the README. 
For devsite, the idea is to point to this section and/or to the doi directly, or we can even duplicate the information, it is not too much and is not expected to change much (at all).

95-martin-orion (approved):

PR #210, opened 2020-10-16
Update Cirq gates.

sergeisakov:
Fix pi constant (M_PI is not portable). Fix 2-qubit matrix gates (add matrix reshuffling to conform with the standard qubit ordering). Add more functions to create matrix gates.


95-martin-orion (approved):
Looks good - thanks for the thorough test cases!

PR #213, opened 2020-10-19
Use latest Cirq version in qsim

95-martin-orion:
Thanks to the Cirq backwards-compatibility policies, we can safely use the latest version of Cirq as long as any Cirq deprecations are reflected in qsim before the next Cirq release (~3 months).


sergeisakov (approved):

balopat (approved):
LGTM

PR #214, opened 2020-10-20
Fix gates with nonzero global shift.

sergeisakov:
This PR fixes gates that have "magic" exponents and nonzero global shifts.

### DIFF
--- a/qsimcirq/qsim_circuit.py
+++ b/qsimcirq/qsim_circuit.py
@@ -29,33 +29,34 @@ def _cirq_gate_kind(gate):
       + 'only 1- or 2-qubit gates are supported.')
   if isinstance(gate, cirq.ops.XPowGate):
     # cirq.rx also uses this path.
-    if gate.exponent == 1:
+    if gate.exponent == 1 and gate.global_shift == 0:
       return qsim.kX
     return qsim.kXPowGate
   if isinstance(gate, cirq.ops.YPowGate):
     # cirq.ry also uses this path.
-    if gate.exponent == 1:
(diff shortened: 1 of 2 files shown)


balopat (changes requested):
Hi Sergei, I just saw this and I figured I jump in - thank you for creating this, this seems like a pretty important fix - can you please add unit tests to cover the failing cases?

qsimcirq/qsim_circuit.py
```diff
@@ -29,58 +29,59 @@ def _cirq_gate_kind(gate):
       + 'only 1- or 2-qubit gates are supported.')
   if isinstance(gate, cirq.ops.XPowGate):
     # cirq.rx also uses this path.
-    if gate.exponent == 1:
+    if gate.exponent == 1 and gate.global_shift == 0:
```
95-martin-orion, line 32: To simplify this, we could instead add the following block at the start of this function:
```
if isinstance(gate, cirq.EigenGate) and gate.global_shift != 0:
  return NotImplementedError(f'qsim does not support global shift.
      Received gate with global shift of {gate.global_shift}.')
```
(`global_shift` is a property of the `EigenGate` class)

95-martin-orion: I see - I misunderstood. This should be good to go, then.

95-martin-orion (approved):

sergeisakov:
> Hi Sergei, I just saw this and I figured I jump in - thank you for creating this, this seems like a pretty important fix - can you please add unit tests to cover the failing cases?

I added some tests. Please check.

balopat (approved):

PR #216, opened 2020-10-21
Add initial support for multi-qubit and controlled gates.

sergeisakov:


lib/fuser.h
```diff
@@ -28,22 +29,21 @@ namespace qsim {
...
   typename Gate::GateKind kind;
   /**
-   * The time index of the first ("master") gate.
+   * The time index of the first ("parent") gate.
    */
   unsigned time;
-  unsigned num_qubits;
   /**
-   * A list of qubits these gates act upon.
+   * A list of qubits these gates act upon, possibly excluding control qubits.
```
95-martin-orion, line 40: Are control qubits only excluded for explicitly-controlled gates (i.e. CX would include its control, but controlled-iSWAP would not)? If so, please note that in the comment here.

lib/gate.h
```diff
@@ -37,43 +58,82 @@ struct Gate {
...
-  unsigned num_qubits;
   std::vector<unsigned> qubits;
+  std::vector<unsigned> controlled_by;
+  uint64_t cmask;
   std::vector<fp_type> params;
   Matrix<fp_type> matrix;
-  bool unfusible;      // If true, the gate is fused as a master.
-  bool inverse;        // If true, the qubit order is inversed (q0 > q1).
+  bool unfusible;      // If true, the gate is fused as a parent.
+  bool swapped;        // If true, the qubit order is inversed (q0 > q1).
```
95-martin-orion, line 67: Does this still indicate inverted qubit order, or is another ordering (e.g. `0, 2, 1` -> `0, 1, 2` for a controlled gate) possible? Please update the comment if so.

lib/gate.h
```diff
@@ -37,43 +58,82 @@ struct Gate {
...
+               std::forward<Matrix<typename Gate::fp_type>>(matrix),
+               false, false};
+
+  detail::SortQubits<Gate, GateDef>(gate);
+
+  return gate;
+}
+
+template <typename Gate>
+inline void MakeControlledGate(std::vector<unsigned>&& controlled_by,
```
95-martin-orion, line 116: The `MakeControlledGate` methods should probably set `symmetric` to false, since adding control qubits might not preserve gate symmetry.

tests/circuit_qsim_parser_test.cc
```diff
@@ -311,6 +329,79 @@ R"(2
...
+0 h 1
+1 m 0 0)";
+
+  std::stringstream ss(invalid_circuit);
+  Circuit<GateQSim<float>> circuit;
+
+  EXPECT_FALSE(CircuitQsimParser<IO>::FromStream(99, provider, ss, circuit));
+}
+
+TEST(CircuitQsimParserTest, NoControllQubits) {
```
95-martin-orion, line 345: Controll -> Control

lib/gate_appl.h
```diff
@@ -35,7 +35,12 @@ template <typename Simulator, typename Gate>
 inline void ApplyGate(const Simulator& simulator, const Gate& gate,
                       typename Simulator::State& state) {
   if (gate.kind != gate::kMeasurement) {
-    simulator.ApplyGate(gate.qubits, gate.matrix.data(), state);
+    if (gate.controlled_by.size() == 0) {
+      simulator.ApplyGate(gate.qubits, gate.matrix.data(), state);
+    } else {
+      simulator.ApplyControlledGate(gate.qubits, gate.controlled_by,
+                                    gate.cmask, gate.matrix.data(), state);
+    }
```
95-martin-orion, line 43: Would it be worthwhile for `ApplyControlledGate` to default to `ApplyGate` if there are no control qubits?

95-martin-orion (changes requested):
First pass. Still need to review `circuit_qsim_parser[_test]` and `fuser_basic_test`.

lib/fuser.h
sergeisakov: Done.

lib/gate.h
sergeisakov: Done.

lib/gate.h
sergeisakov: `symmetric` is only used to check if the gate matrix needs to be reshuffled if the gate qubits are not ordered. Adding control qubits does not change the gate matrix and it preserves gate symmetry.

tests/circuit_qsim_parser_test.cc
sergeisakov: Done.

lib/gate_appl.h
sergeisakov: Do you mean to use just one function, i.e. to merge `ApplyControlledGate` and `ApplyGate`?

95-martin-orion: Either that, or conditionally calling `ApplyGate` at the start of `ApplyControlledGate`:
```
void ApplyControlledGate (...) {
  if (gate.controlled_by.size() == 0) {
    ApplyGate(...);
    return;
  }
  /* ... rest of function ... */
}
```

tests/circuit_qsim_parser_test.cc
```diff
@@ -70,6 +70,24 @@ R"(2
   EXPECT_EQ(circuit.gates.size(), 10);
 }
 
+TEST(CircuitQsimParserTest, ValidCircuitWithControlledGates) {
+  constexpr char valid_circuit[] =
+R"(5
+0 c 0 1 h 2
```
95-martin-orion, line 76: I don't think this is a valid circuit. A qubit can't simultaneously control one gate and be acted on by another, as is the case with qubits {0, 1, 2} in this example.

tests/fuser_basic_test.cc
```diff
@@ -72,152 +72,152 @@ TEST(FuserBasicTest, NoTimesToSplitAt) {
 
   EXPECT_EQ(fused_gates[0].kind, kGateCZ);
   EXPECT_EQ(fused_gates[0].time, 1);
-  EXPECT_EQ(fused_gates[0].num_qubits, 2);
+  EXPECT_EQ(fused_gates[0].qubits.size(), 2);
```
95-martin-orion, line 75: Suggestion: with the amount of repetition in this file, it seems like it would be worthwhile to define equality operators for gates and fused gates. Doesn't need to be a part of this PR, though.

lib/gate.h
```diff
@@ -37,56 +65,95 @@ struct Gate {
 
   GateKind kind;
   unsigned time;
-  unsigned num_qubits;
   std::vector<unsigned> qubits;
+  std::vector<unsigned> controlled_by;
+  uint64_t cmask;
```
95-martin-orion, line 70: Does this mask specify the required state of the control qubits to apply the gate? i.e. for `controlled_by = {0, 1, 2}; cmask = 0b101`, would the gate trigger for state `|101)`?

95-martin-orion (changes requested):
All files reviewed.

lib/gate_appl.h
sergeisakov: Yes, that makes sense. I'll use the second option.

lib/circuit_qsim_parser.h
```diff
@@ -265,37 +167,228 @@ class CircuitQsimParser final {
...
-   * @param line Line number of the parsed gate; only used for error reporting.
    */
   static bool ValidateGate(std::stringstream& ss, unsigned num_qubits,
-                           const std::vector<unsigned>& qubits,
-                           const std::string& provider, unsigned line) {
-    for (auto q : qubits) {
-      if (q >= num_qubits) {
-        InvalidGateError(provider, line);
+                           const std::vector<unsigned>& qubits) {
+    return ss && /*ss.peek() != std::stringstream::traits_type::eof();
```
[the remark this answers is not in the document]
sergeisakov: Done.

tests/circuit_qsim_parser_test.cc
sergeisakov: Done.

tests/fuser_basic_test.cc
sergeisakov: Acknowledged.

lib/gate.h
sergeisakov: Yes, exactly.

tests/circuit_qsim_parser_test.cc
95-martin-orion: The fact that the test previously passed suggests that this is still supported in the parser. Could you update the parser and add a test to make sure that simultaneous control and action on a qubit is rejected?

95-martin-orion (changes requested):

tests/circuit_qsim_parser_test.cc
sergeisakov: Simultaneous control and action on a qubit is rejected, see lines 381 and 393 in this files. You mean that it is possible to have gates with equal times (or moments in Cirq language) that act on the same qubits, right? This is true for all the gates not only for controlled gates. So perhaps that should be fixed by a separate PR.

95-martin-orion: Ah, I wasn't aware that was the case. In that case it's fine to leave this to a separate PR - I'll open an issue to track it.

95-martin-orion (approved):

PR #219, opened 2020-10-22
Polish qsimcirq getting started guide.

rmlarose:
Removes unused `states` variable, standardizes package installation as per https://github.com/quantumlib/Cirq/pull/3431, and small formatting changes.


95-martin-orion (approved):
Thanks for taking a look at these!

PR #224, opened 2020-10-27
Add overview to subsite tabs

95-martin-orion:


sergeisakov (approved):

docs/_book.yaml
```diff
@@ -10,6 +10,8 @@ upper_tabs:
     other:
     - name: "Guide & Tutorials"
       contents:
+      - title: "Overview"
```
lamberta, line 13: Can we name this "qsim and qsimh" to avoid a naming conflict with the "Overview" tab above?

95-martin-orion: We can, although I don't think we can change the path (`/qsim/overview`) to match. Is that OK?

lamberta: Yeah, that's fine. Can leave the filename

95-martin-orion: Done.

lamberta (approved):

PR #225, opened 2020-10-27
Add multi-qubit and multi-control gates.

sergeisakov:


tests/simulator_testfixture.h
```diff
@@ -46,12 +48,19 @@ void TestApplyGate1() {
   auto gate5 = GateZ<fp_type>::Create(4, 0);
   auto gate6 = GateHZ2<fp_type>::Create(5, 0);
 
+  puts("obs1");
```
95-martin-orion, line 51: Remove debug lines?

lib/simulator_avx.h
```diff
@@ -19,15 +19,15 @@
...
 
+#include "bits.h"
 #include "statespace_avx.h"
 
 namespace qsim {
 
 /**
- * Quantum circuit simulator with AVX vectorization.
- */
+* Quantum circuit simulator with AVX vectorization.
```
95-martin-orion, line 29: Top-level comment: if this is generated code, could you include the script used to generate it in the PR? That would be easier to review, and provides a reference for how these changes were made.

95-martin-orion (changes requested):

tests/simulator_testfixture.h
sergeisakov: Done.

lib/simulator_avx.h
sergeisakov: I would prefer to not include the script at this point.

95-martin-orion (approved):

lib/simulator_avx.h
95-martin-orion: Discussed offline: submitting the script to the repo is lower priority. #227 has been opened to track that work.

PR #228, opened 2020-10-29
Update simulator tests.

sergeisakov:


tests/simulator_testfixture.h
```diff
@@ -346,123 +345,159 @@ void TestApplyGate5() {
...
+
+  gates.push_back(GateHd<fp_type>::Create(0, 0));
+  gates.push_back(GateHd<fp_type>::Create(0, 1));
+  gates.push_back(GateHd<fp_type>::Create(0, 2));
+  gates.push_back(GateHd<fp_type>::Create(0, 3));
+  gates.push_back(GateHd<fp_type>::Create(0, 4));
+  gates.push_back(GateHd<fp_type>::Create(0, 5));
+
+  gates.push_back(GateT<fp_type>::Create(1, 1));
+  MakeControlledGate({0}, {1}, gates.back());
```
95-martin-orion, line 365: If `MakeControlledGate` returned the generated gate, this sequence of `push_back`s could be replaced with one big vector-initialization, which would parallel nicely with Cirq's Circuit constructors.

95-martin-orion (approved):
Looking good - just one suggestion before merging.

tests/simulator_testfixture.h
sergeisakov: This is a good point.

PR #229, opened 2020-10-30
Update docs landing page.

lilies:
Includes content and styling updates.

### DIFF
--- a/docs/_index.yaml
+++ b/docs/_index.yaml
@@ -19,55 +19,101 @@ description: >
 landing_page:
   custom_css_path: /site-assets/css/style.css
   rows:
-  - heading: Quantum circuit simulators qsim and qsimh.
-
+  - classname: quantum-hero quantum-hero--qsim quantum-hero--icon-medium
+    options:
+    - hero
+    - description-50
+    - padding-large
+    heading: qsim and qsimh
+    icon:
+      path: /site-assets/images/icons/icon_qsim.png
+    description: >
+      Optimized quantum circuit simulators
     items:
-    - classname: devsite-landing-row-50
+    - description: >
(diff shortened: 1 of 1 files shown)


docs/_index.yaml
```diff
@@ -19,55 +19,101 @@ description: >
...
+        path: //github.com/quantumlib/qsim
+        classname: button button-white
+        icon_name: open_in_new
+      code_block: |
+        <pre class = "prettyprint">
+        import cirq
+        import qsimcirq
+
+        # Pick up to ~36 qubits to run on a large multi-core Intel CPU machine
+        qubits = [cirq.GridQubit(i,j) for i in range(6) for j in range(6)]
```
95-martin-orion, line 63: 36 qubits is pretty ambitious for an example - even on a powerful machine, it may take several minutes to run. Another consideration: for general-purpose Colab machines, ~28 qubits is the maximum that fits in memory.

If we want to keep the large example, maybe add a line recommending smaller qubit counts for test runs?

docs/_index.yaml
```diff
@@ -19,55 +19,101 @@ description: >
...
+
+        # Define a circuit to run (e.g. Quantum Supremacy Circuits)
```
95-martin-orion, line 65: Prefer "Beyond Classical" in place of "Quantum Supremacy". (The only exception is when explicitly referring to the 2019 Google experiment - if that's the case here, please clarify.)

95-martin-orion (changes requested):
Is there a generated preview of this page that I can view? It's difficult to review e.g. links and images in this format.

95-martin-orion (approved):
Discussed offline - all changes here have been pre-approved.

PR #230, opened 2020-10-30
Fix unitary calculator.

### LINKED ISSUE #226: Unitary classes have no State type
@MichaelBroughton 

As of PR #188, `ApplyFusedGate` expects the calling Simulator class to define a `State` type for one of its arguments. UnitaryCalculatorBasic and UnitaryStateSpace have such an object, but name it `Unitary` instead. We need to fix the naming and add tests that verify `ApplyFusedGate` works for the Unitary classes.

sergeisakov:
Closes #226


MichaelBroughton (approved):
LGTM. Thanks for jumping on this!

PR #232, opened 2020-11-03
Add options for gate fusion.

sergeisakov:


lib/fuser_basic.h
```diff
@@ -31,26 +31,35 @@ template <typename IO, typename Gate>
 struct BasicGateFuser final {
   using GateFused = qsim::GateFused<Gate>;
 
+  /**
+   * User-specified parameters for gate fusion.
+   * BasicGateFuser does not use any parameters.
+   */
+  struct Parameter {};
```
95-martin-orion, line 38: I assume fusion with more than two qubits will be handled by a new Fuser type?

95-martin-orion (approved):
My reading of this is that users will pass Fuser params into run_qsim calls, and those params will be consumed by a new Fuser implementation, to be defined in a follow-up PR.

Is that correct?

sergeisakov:
> My reading of this is that users will pass Fuser params into run_qsim calls, and those params will be consumed by a new Fuser implementation, to be defined in a follow-up PR.
> 
> Is that correct?

Yes, that is correct.

lib/fuser_basic.h
sergeisakov: Yes, exactly.

PR #234, opened 2020-11-06
Add 3-qubit Cirq gates.

sergeisakov:
This PR also adds multi-qubit identity and matrix gates.

PR #236, opened 2020-11-09
Fallback on unitary() in gate decomposition.

95-martin-orion:
Follow-up to #234. This PR ensures that 3+ qubit gates with no path for decomposing into known 1- and 2-qubit gates can be handled by qsim by converting to MatrixGates.

### DIFF
--- a/qsimcirq/qsim_circuit.py
+++ b/qsimcirq/qsim_circuit.py
@@ -187,11 +187,21 @@ class QSimCircuit(cirq.Circuit):
     # qsim numbers qubits in reverse order from cirq
     ordered_qubits = list(reversed(ordered_qubits))
 
+    def has_qsim_kind(op: cirq.ops.GateOperation):
+      return _cirq_gate_kind(op.gate) != None
+
+    def to_matrix(op: cirq.ops.GateOperation):
+      mat = cirq.protocols.unitary(op.gate, None)
+      if mat is None:
+          return NotImplemented
+      
+      return cirq.ops.MatrixGate(mat).on(*op.qubits)
+
     qubit_to_index_dict = {q: i for i, q in enumerate(ordered_qubits)}
     time_offset = 0
     for moment in self:
       ops_by_gate = [
-        cirq.decompose(op, keep=lambda x: _cirq_gate_kind(x.gate) != None)
+        cirq.decompose(op, fallback_decomposer=to_matrix, keep=has_qsim_kind)
         for op in moment
       ]
       moment_length = max(len(gate_ops) for gate_ops in ops_by_gate)
--- a/qsimcirq_tests/qsimcirq_test.py
+++ b/qsimcirq_tests/qsimcirq_test.py
@@ -228,14 +228,17 @@ class MainTest(unittest.TestCase):
       [1, 0, 0, 0, 0, 0, 0, 0],
       [0, 1, 0, 0, 0, 0, 0, 0],
       [0, 0, 1, 0, 0, 0, 0, 0],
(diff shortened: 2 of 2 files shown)


qsimcirq/qsim_circuit.py
```diff
@@ -187,11 +187,21 @@ def translate_cirq_to_qsim(
     # qsim numbers qubits in reverse order from cirq
     ordered_qubits = list(reversed(ordered_qubits))
 
+    def has_qsim_kind(op: cirq.ops.GateOperation):
+      return _cirq_gate_kind(op.gate) != None
```
sergeisakov, line 191: [No action required]  `_cirq_gate_kind` is called here and below (line 215). Would it be possible to call it only once?

qsimcirq_tests/qsimcirq_test.py
```diff
@@ -221,6 +221,70 @@ def test_matrix2_gate(self):
         result.state_vector(), cirq_result.state_vector())
 
 
+  def test_big_matrix_gates(self):
+    qubits = cirq.LineQubit.range(3)
+    # Toffoli gate as a matrix.
+    m = np.array([
```
sergeisakov, line 227: I think the Toffoli matrix here and below should be in the standard form. One thing is that this test tests only the first column of the matrix.

sergeisakov (changes requested):

qsimcirq/qsim_circuit.py
95-martin-orion: This is just explicitly defining the `lambda` that was previously in the `cirq.decompose` call. Due to the required return types for `cirq.decompose`, it can't really be re-written to fit both use cases.

qsimcirq_tests/qsimcirq_test.py
95-martin-orion: Done. Also modified the circuit to ensure the Toffolis aren't no-ops.

I'm not certain what you mean by only testing the first column of the matrix - this test should execute the circuit (Toffolis included) and verify that the resulting state matches the result from Cirq.

qsimcirq/qsim_circuit.py
sergeisakov: I think that, for each gate, `_cirq_gate_kind` is called in line 204 (implicitly via the `lambda` function) and one more time in line 215. I was wondering if it is possible to modify the code in such a way that `_cirq_gate_kind` is called only once for each gate.

sergeisakov (approved):

qsimcirq/qsim_circuit.py
95-martin-orion: Ah, sorry for the confusion. `cirq.decompose` mutates the circuit and may call `_cirq_gate_kind` multiple times for each gate. There may be a solution involving memoization, but it would require changes to `cirq.decompose` which aren't really feasible here. Thanks for checking, though!

qsimcirq_tests/qsimcirq_test.py
sergeisakov: The starting state vector is [1, 0, 0, 0, 0, 0, 0, 0]. Only the first column of the matrix is relevant in the matrix-vector product in this case.

95-martin-orion: Ah, I see. Added Hadamards on the control qubits for a more complete test of the matrix.

The main failure case we're concerned about here is being unable to decompose, but I agree it's better to fully test the matrix being used (e.g. if it was translated incorrectly somehow).

PR #243, opened 2020-11-20
Simplified GCP without token

jrossthomson:
Removed token requirement for connection. This simplifies the connection procedure. There is a security issue if this is not a new project: if the user has already created a firewall rule to expose port 8888 the notebook server is exposed. The notebook has access to shell.

Removing token auth is probably not recommended. I'll leave it to @95-martin-orion to decide if this PR is worth it to trade simplicity for security.


95-martin-orion:
I'm not familiar enough with Google policy surrounding OSS security requirements to approve this PR. If we want to push forward with this, I think the correct course of action would be to first get confirmation from the relevant internal teams.

jrossthomson:
I'm going to close this for now. There may be a better way to ensure a simpler secure connection, but I don't think this is it.

PR #245, opened 2020-11-24
Add multi-qubit fuser.

sergeisakov:


apps/qsim_amplitudes.cc
```diff
@@ -172,9 +178,9 @@ int main(int argc, char* argv[]) {
...
   };
 
-  using Runner = QSimRunner<IO, BasicGateFuser<IO, GateQSim<float>>, Simulator>;
-
   Runner::Parameter param;
+  param.max_fused_size = opt.max_fused_size;
+  param.fuser_verbosity = opt.verbosity;
   param.seed = opt.seed;
   param.num_threads = opt.num_threads;
   param.verbosity = opt.verbosity;
```
95-martin-orion, line 186: Is `param.verbosity` still in use separate from `param.fuser_verbosity`, or can it be removed?

Similarly in the other binary files.

docs/usage.md
```diff
@@ -11,14 +11,15 @@ Sample circuits are provided in [circuits](/circuits).
...
 ```
 
-| Flag | Description | 
+| Flag | Description |
 |-------|------------|
-|`-c circuit_file` | circuit file to run| 
+|`-c circuit_file` | circuit file to run|
 |`-d maxtime` | maximum time |
 |`-t num_threads` | number of threads to use|
+|`-f max_fused_size` | maximum fused gate size|
```
95-martin-orion, line 22: We should include a description somewhere (perhaps in the tutorial notebook) which explains what this flag does.

tests/fuser_mqubit_test.cc
```diff
@@ -0,0 +1,295 @@
...
+    while (n > max_fused_gate_size) {
+      AddToCircuit(time, distr, rgen, n, available, circuit);
+    }
+  }
+
+  return circuit;
+}
+
+template <typename FusedGate>
+bool TestFusedGates(unsigned num_qubits,
```
95-martin-orion, line 176: If I understand it right, this will verify that the circuit in question is valid (no timing conflicts, measurements are isolated, and the input gates are only in the fused result once).

Could you add a test verifying that a circuit with 3+ qubit fusing can be run (e.g. with `run_qsim`) and that the results match those of a circuit run with basic fusion?

95-martin-orion (changes requested):
Read through all but `fuser_mqubit.h` - will cover that in my next pass.

apps/qsim_amplitudes.cc
sergeisakov: I removed `param.fuser_verbosity`.

docs/usage.md
sergeisakov: Yes, we should.

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+
+    unsigned cur_size = fgate.qubits.size();
+    fgate.visited = level;
+
+    unsigned max_size = cur_size;
+
+    GetNextAvailableGates(max_fused_size, cur_size, fgate, nullptr,
+                          scratch.data, scratch.next1);
+
+    for (auto n1 : scratch.next1) {
```
95-martin-orion, line 702: These for-loops reuse a lot of the same code. Could this be restructured as a recursive function call without negatively affecting performance?

tests/fuser_mqubit_test.cc
sergeisakov: Yes, that is right. I added a test that uses `ApplyGate` (no gate fusion) and `ApplyFusedGate` (gate fusion). I did not use basic fusion as it might get deprecated at some point.

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+
+      auto pgate = link->prev->val;
+
+      if (pgate->visited == 1) {
+        MakeGateSequence(max_fused_size, scratch, *pgate);
+      }
+    }
+  }
+
+  static void FindLongestGateSequence(
```
95-martin-orion, line 684: After reading this, I think I understand the diagrams above a bit better. My reading is that this will probe both forwards and backwards in time from `fgate` to find the longest gate sequence (on no more than `max_fused_size` qubits) that it can be a part of. Is that correct?

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+      unsigned cur_size2 = cur_size + n1->qubits.size();
+      if (cur_size2 > max_fused_size) continue;
+
+      bool feasible = GetPrevAvailableGates(max_fused_size, cur_size,
+                                            level, *n1->gate, nullptr,
+                                            scratch.data, scratch.prev1);
+
+      if (!feasible) continue;
+
+      if (scratch.prev1.size() == 0 && max_fused_size > 3) continue;
```
95-martin-orion, line 712: This looks like it enforces a minimum probe depth. Why does this only appear at the top level, and why is.3 qubits the threshold?

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+  }
+
+  static void GetAddedQubits(const GateF& fgate0, const GateF* fgate1,
+                             const GateF& fgate2, GateA& added) {
+    for (std::size_t i = 0; i < fgate2.qubits.size(); ++i) {
+      unsigned q2 = fgate2.qubits[i];
+
+      bool have_added_qubits = true;
+
+      for (auto q0 : fgate0.qubits) {
```
95-martin-orion, line 897: Using `continue` would be cleaner than a flag here, I think. For example:
```
if (std::find(q2, fgate0.qubits.begin(), fgate0.qubits.end())
    == fgate0.qubits.end()) {
  continue;
}
```

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+
+    bool may_have_gates_to_fuse = true;
+
+    while (may_have_gates_to_fuse) {
+      may_have_gates_to_fuse = false;
+
+      auto time = fgate.parent->time;
+
+      std::sort(links.begin(), links.end(),
+                [&time](const Link* l, const Link* r) -> bool {
```
95-martin-orion, line 980: `time` is not used in the function body.

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+    uint64_t bad_mask = 0;
+    auto links = fgate.links;
+
+    bool may_have_gates_to_fuse = true;
+
+    while (may_have_gates_to_fuse) {
+      may_have_gates_to_fuse = false;
+
+      std::sort(links.begin(), links.end(),
+                [](const Link* l, const Link* r) -> bool {
```
95-martin-orion, line 934: This lambda is reused in FuseNext - it can probably be separated out as a helper method.

Alternatively, it seems like FusePrev and FuseNext are very similar - the shared portion may be possible to extract into a helper method.

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+    }
+  }
+
+  static bool QubitsAreIn(uint64_t mask0, uint64_t mask) {
+    return ((mask0 | mask) ^ mask0) == 0;
+  }
+
+  static void PrintStat(unsigned verbosity, const Stat& stat,
+                        const std::vector<GateFused>& fused_gates) {
+    if (verbosity > 0) {
```
95-martin-orion, line 1032: To reduce indentation, I recommend `if (verbosity <= 0) return;` instead.

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+          } else {
+            IO::messagef(", ");
+          }
+          IO::messagef("%u %u-qubit", stat.num_gates[i], i);
+        }
+      }
+
+      IO::messagef(" gates are fused into %lu gates\n", stat.num_fused_gates);
+
+      if (verbosity > 1) {
```
95-martin-orion, line 1060: Similarly, `if (verbosity <= 1) return;` here.

95-martin-orion (changes requested):
Comments and questions on `fuser_mqubit.h`.

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+    const Gate* parent;
+    std::vector<unsigned> qubits;
+    std::vector<const Gate*> gates; // Gates that get fused to this gate.
+    std::vector<Link*> links;       // Gate "lattice" links.
+    uint64_t mask;                  // Qubit mask.
+    unsigned visited;
+  };
+
+  using LinkManager = LinkManagerT<GateF*>;
+  using Link = typename LinkManager::Link;
```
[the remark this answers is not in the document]
sergeisakov: Thank you for the review. Done.

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+ public:
+  using GateFused = qsim::GateFused<Gate>;
+
+  /**
+   * User-specified parameters for gate fusion.
+   */
+  struct Parameter {
+    /**
+     * Maximum number of qubits in a fused gate. It can take values from 2 to
+     * 6 (0 and 1 are equivalent to 2). It is not recommended to use 5 or 6.
```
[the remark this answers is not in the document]
sergeisakov: Yes, that is correct. I added a comment.

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+        prev_time = gate.time;
+
+        // Fill in auxillary data structures.
+
+        if (gate.kind == gate::kMeasurement) {
+          // Measurement gate.
+
+          if (last_mea_gate == nullptr
+              || last_mea_gate->parent->time != gate.time) {
+            gates_seq.push_back({&gate, {}, {}, {}, 0, 997});
```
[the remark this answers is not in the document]
sergeisakov: Done.

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+        FusePrev(0, *fgate);
+        fgate->gates.push_back(fgate->parent);
+        FuseNext(0, *fgate);
+      }
+
+      fgates[max_gate_size - i].resize(pos);
+    }
+  }
+
+  // Try to fuse gate sequences as follows. Gate time goes from left to right.
```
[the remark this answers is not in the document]
sergeisakov: Done.

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+  //
+  // max_fused_size = 3:
+  //  |
+  //   |
+  //
+  // max_fused_size = 4:
+  //  |
+  //   |
+  //  |
```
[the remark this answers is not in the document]
sergeisakov: Start with some gate. Then go forward in time. Then backward and so on.
A line in the picture represents a (two-qubit) gate. max_fused_size is the maximum number of qubits in a fused gate (gate size).

lib/fuser_mqubit.h
```diff
@@ -0,0 +1,1087 @@
...
+  //   |
+  //  |
+  //
+  static void FuseGateSequences(unsigned max_fused_size, unsigned num_qubits,
+                                std::vector<GateF>& gates_seq, Stat& stat,
+                                std::vector<GateFused>& fused_gates) {
+    Scratch scratch;
```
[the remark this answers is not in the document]
sergeisakov: Done. Though it affects performance only in a limited number of cases.

lib/fuser_mqubit.h
sergeisakov: Removed some code duplication.

lib/fuser_mqubit.h
sergeisakov: Yes, that is correct.

lib/fuser_mqubit.h
sergeisakov: The result can be just _- for max_fused_size > 3. It turns out that it is better to have just _ in this case.

lib/fuser_mqubit.h
sergeisakov: Done.

lib/fuser_mqubit.h
sergeisakov: Removed.

lib/fuser_mqubit.h
sergeisakov: Added a helper method.

lib/fuser_mqubit.h
sergeisakov: Done.

lib/fuser_mqubit.h
sergeisakov: Done.

lib/fuser_mqubit.h
```diff
@@ -85,8 +89,19 @@ class MultiQubitGateFuser {
...
   };
 
-  using LinkManager = LinkManagerT<GateF*>;
-  using Link = typename LinkManager::Link;
+  // Possible values for visited in GateF.
+  // Note that MakeGateSequence assignes values from kSecond to the number of
+  // gates in the sequence plus one, see below.
+  enum Visited {
+    kZero = 0,             // Start value for "normal" gates.
+    kFirst = 1,            // Valut after the first pass for partially fused
```
95-martin-orion, line 97: Valut -> Value

lib/fuser_mqubit.h
```diff
@@ -85,8 +89,19 @@ class MultiQubitGateFuser {
     unsigned visited;
```
95-martin-orion, line 89: `GateF.visited` can be of type `Visited`, just to make that relationship explicit.

95-martin-orion (approved):
The new test looks good! A couple of leftover nits, but nothing blocking.

lib/fuser_mqubit.h
sergeisakov: Done.

lib/fuser_mqubit.h
sergeisakov: Acknowledged.

PR #246, opened 2020-12-02
Pass through max fused qubits from Python

95-martin-orion:
Allows Python users to pipe in the `max_fused_size` parameter with option `f`.

### DIFF
--- a/docs/tutorials/qsimcirq.ipynb
+++ b/docs/tutorials/qsimcirq.ipynb
@@ -376,6 +376,33 @@
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


sergeisakov (approved):
It would be good to document this option somewhere (say, in docs/cirq_interface.md or in docs/tutorials/qsimcirq.ipynb).

95-martin-orion:
> It would be good to document this option somewhere (say, in docs/cirq_interface.md or in docs/tutorials/qsimcirq.ipynb).

Done.

Issue #227, opened 2020-10-28
Add code-generating scripts to the repo


95-martin-orion:
PR #225 was partially generated by a script which is not currently hosted on Github. For future reference and maintenance purposes, we should include that script in the qsim repo.

This issue is low priority relative to the ongoing multi-qubit work.

95-martin-orion:
Much of the code in #248 is also generated from a script, which we should also include in the repo.

mhucka:
This is old enough that I wonder whether the scripts can even be located anymore. I'm going to close this due to age and low priority. If that's incorrect, someone please feel free to reopen this.

PR #248, opened 2020-12-03
Add API for computing expectation values.

### LINKED ISSUE #238: Provide an API for finding expectation values
Related to [Cirq issue 3492](https://github.com/quantumlib/Cirq/issues/3492). This also provides partial relief for issue #237.

Similar to the existing API for state vector sampling, qsim should provide an API for returning expectation values from a simulation.

sergeisakov:
Closes #238

### DIFF
--- a/lib/expect.h
+++ b/lib/expect.h
@@ -1,0 +1,126 @@
+// Copyright 2019 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef EXPECT_H_
+#define EXPECT_H_
+
+#include <complex>
+
+#include "fuser.h"
+#include "gate_appl.h"
+
+namespace qsim {
+
+template <typename Gate>
+struct OpString {
+  std::complex<double> weight;
+  std::vector<Gate> ops;
+};
+
+/**
+ * Computes the expectation value of the sum of operator strings. Operators
+ * can be any supported gates. Uses a temporary state vector.
+ * @param param Options for gate fusion.
+ * @param strings Operator strings.
+ * @param ss StateSpace object required to copy the state vector and compute
+ *   inner products.
+ * @param simulator Simulator object. Provides specific implementations for
+ *   applying gates.
+ * @param state The state vector of the system.
+ * @param state Temporary state vector.
+ * @return The computed expectation value.
+ */
+template <typename Fuser, typename Simulator, typename Gate>
+std::complex<double> ExpectaionValue(const typename Fuser::Parameter& param,
+                                     const std::vector<OpString<Gate>>& strings,
+                                     const typename Simulator::StateSpace& ss,
+                                     const Simulator& simulator,
+                                     const typename Simulator::State& state,
+                                     typename Simulator::State& ket) {
+  std::complex<double> eval = 0;
+
+  for (const auto& str : strings) {
+    if (str.ops.size() == 0) continue;
+
+    ss.Copy(state, ket);
+
+    if (str.ops.size() == 1) {
+      const auto& op = str.ops[0];
+      simulator.ApplyGate(op.qubits, op.matrix.data(), ket);
+    } else {
+      auto fused_gates = Fuser::FuseGates(param, state.num_qubits(), str.ops);
+      if (fused_gates.size() == 0) {
+        eval = 0;
+        break;
+      }
+
+      for (const auto& fgate : fused_gates) {
+        ApplyFusedGate(simulator, fgate, ket);
+      }
+    }
+
+    eval += str.weight * ss.InnerProduct(state, ket);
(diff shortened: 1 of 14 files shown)


lib/simulator_basic.h
```diff
@@ -853,6 +901,457 @@ class SimulatorBasic final {
              state.num_qubits(), cmaskh, emaskh, rstate);
   }
 
+  std::complex<double> ExpectationValue1H(const std::vector<unsigned>& qs,
```
95-martin-orion, line 904: These `ExpectationValue*H` methods are very similar to each other. Would combining them negatively impact performance? (Integer template values may help with this)

lib/expect.h
```diff
@@ -0,0 +1,126 @@
...
+ * @return The computed expectation value.
+ */
+template <typename Fuser, typename Simulator, typename Gate>
+std::complex<double> ExpectaionValue(const std::vector<OpString<Gate>>& strings,
+                                     const Simulator& simulator,
+                                     const typename Simulator::State& state) {
+  std::complex<double> eval = 0;
+
+  typename Fuser::Parameter param;
+  param.max_fused_size = 6;
```
95-martin-orion, line 98: Would calculating `max_fused_size` from the number of qubits in `strings` be worthwhile here?

lib/expect.h
```diff
@@ -0,0 +1,126 @@
...
+ * Computes the expectation value of the sum of operator strings. Operators
+ * can be any supported gates. Uses a temporary state vector.
+ * @param param Options for gate fusion.
+ * @param strings Operator strings.
+ * @param ss StateSpace object required to copy the state vector and compute
+ *   inner products.
+ * @param simulator Simulator object. Provides specific implementations for
+ *   applying gates.
+ * @param state The state vector of the system.
+ * @param state Temporary state vector.
```
95-martin-orion, line 41: `state` -> `ket` for this argument (`state` is on the previous line)

lib/expect.h
```diff
@@ -0,0 +1,126 @@
...
+ * @return The computed expectation value.
+ */
+template <typename Fuser, typename Simulator, typename Gate>
+std::complex<double> ExpectaionValue(const typename Fuser::Parameter& param,
```
95-martin-orion, line 45: `ExpectaionValue` -> `ExpectationValue`

Same elsewhere in this PR.

lib/expect.h
```diff
@@ -0,0 +1,126 @@
...
+struct OpString {
+  std::complex<double> weight;
+  std::vector<Gate> ops;
+};
+
+/**
+ * Computes the expectation value of the sum of operator strings. Operators
+ * can be any supported gates. Uses a temporary state vector.
+ * @param param Options for gate fusion.
+ * @param strings Operator strings.
```
95-martin-orion, line 35: What is an "operator string" in this context? Could we include this info in the docstring? (Same for the other overload of this method)

95-martin-orion (changes requested):
First pass: `expect.h` and `simulator_basic.h`. I'll go over AVX/SSE as well, but I expect my comments will be similar.

lib/simulator_basic.h
```diff
@@ -107,6 +108,53 @@ class SimulatorBasic final {
     }
   }
 
+  /**
+   * Computes the expectation value of an operator using non-vectorized
+  *  instructions.
```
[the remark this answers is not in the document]
sergeisakov: Done.

lib/simulator_basic.h
```diff
@@ -107,6 +108,53 @@ class SimulatorBasic final {
...
+      return ExpectationValue4H(qs, matrix, state);
+      break;
+    case 5:
+      return ExpectationValue5H(qs, matrix, state);
+      break;
+    case 6:
+      return ExpectationValue6H(qs, matrix, state);
+      break;
+    default:
+      // Not implemented.
```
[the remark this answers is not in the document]
sergeisakov: Yes, this case is implemented in expect.h. Functions in Simulator* (and in many other places in qsim) do not validate input parameters. I added raising an error in expect.h.

lib/simulator_basic.h
```diff
@@ -107,6 +108,53 @@ class SimulatorBasic final {
...
+      break;
+    }
+
+    return 0;
+  }
+
+  /**
+   * @return The size of SIMD register if applicable.
+   */
+  unsigned SIMDRegisterSize() {
```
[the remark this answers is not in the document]
sergeisakov: Yes, this method is not used. It might be useful if we add some heuristic to choose the max_fused_size parameter.

lib/simulator_basic.h
sergeisakov: Combining them by using a template with integer template parameters should not impact performance.
However, these methods are generated. So I think it is not a problem if there is a lot of similar code. Anyway combining ExpectationValue* (and ApplyGate*) is a quite involved task and it should not be a part of this PR.

lib/expect.h
sergeisakov: It does not affect anything if `max_fused_size` is greater than the actual number of qubits.

lib/expect.h
sergeisakov: Done.

lib/expect.h
sergeisakov: Done.

lib/expect.h
sergeisakov: It is just a sequence of operators. I updated the docstring.

95-martin-orion:
Having skimmed through the AVX and SSE changes, I can see how templating the ExpectationValue* methods could easily become complicated. Instead, I've updated #227 with a TODO to (eventually) publish the script that generated this code.

95-martin-orion (approved):

PR #253, opened 2020-12-07
Fix sampling in qsimcirq.

sergeisakov:
Makes sampling much faster. The Cirq sample_state_vector() is still quite slow and it should be replaced by a C++ function at some point.


95-martin-orion (approved):
Looks good. For some reason I thought we were already sampling on the C++ side, but I guess that method only does single samples.

PR #254, opened 2020-12-08
Update tests.

sergeisakov:
Updates one simulator test to cover all ApplyGate* functions. Adds a multi-qubit fuser test for small circuits.

### DIFF
--- a/tests/fuser_mqubit_test.cc
+++ b/tests/fuser_mqubit_test.cc
@@ -62,8 +62,8 @@ DummyGate CreateDummyMeasurementGate(unsigned time,
 DummyGate CreateDummyControlledGate(unsigned time,
                                     std::vector<unsigned>&& qubits,
                                     std::vector<unsigned>&& controlled_by) {
-  return
-  {kGateOther, time, 0, std::move(qubits), std::move(controlled_by), {}, false};
+  return {kGateOther, time, 0, std::move(qubits), std::move(controlled_by), {},
+          false};
 }
 
 std::vector<unsigned> GenQubits(unsigned num_qubits, std::mt19937& rgen,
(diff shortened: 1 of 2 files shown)


tests/fuser_mqubit_test.cc
```diff
@@ -371,6 +373,269 @@ TEST(FuserMultiQubitTest, Simulation) {
...
+      CreateDummyGate(2, {5, 6}),
+      CreateDummyGate(3, {3, 4}),
+    };
+
+    {
+      param.max_fused_size = 3;
+      auto fused_gates = Fuser::FuseGates(
+          param, num_qubits, circuit.begin(), circuit.end());
+
+      EXPECT_TRUE(TestFusedGates(num_qubits, circuit, fused_gates));
```
95-martin-orion, line 626: No expectation on `fused_gates.size()` here?

tests/fuser_mqubit_test.cc
```diff
@@ -371,6 +373,269 @@ TEST(FuserMultiQubitTest, Simulation) {
...
+    }
+
+    {
+      param.max_fused_size = 5;
+      auto fused_gates = Fuser::FuseGates(
+          param, num_qubits, circuit.begin(), circuit.end());
+
+      EXPECT_TRUE(TestFusedGates(num_qubits, circuit, fused_gates));
```
95-martin-orion, line 634: No expectation on `fused_gates.size()` here?

95-martin-orion (approved):
The broad range of `fuser_mqubit` tests looks great! A couple of comments on those (not sure if anything needs to be changed) but otherwise this can be merged.

tests/fuser_mqubit_test.cc
sergeisakov: Yes, the reason here and below is that the current version of the fuser does not give the optimal number of fused gates. Perhaps some future version will be better.

PR #255, opened 2020-12-14
Fuse gates referenced by pointers.

sergeisakov:


lib/fuser_basic.h
```diff
@@ -26,10 +26,26 @@ namespace qsim {
...
 struct BasicGateFuser final {
-  using GateFused = qsim::GateFused<Gate>;
+ private:
+  using RGate = typename std::remove_pointer<Gate>::type;
+
+  static const RGate& ConstRefToGate(const RGate& gate) {
+    return gate;
+  }
+
+  static const RGate& ConstRefToGate(const RGate* gate) {
```
95-martin-orion, line 43: Should these be `GateToConstRef` instead? It looks like the output is a const ref, rather than the input.

95-martin-orion (approved):
This all looks good - is there an upcoming feature that this is meant to enable?

sergeisakov:
> This all looks good - is there an upcoming feature that this is meant to enable?

Yes, this will be used in the upcoming quantum trajectory simulator.

lib/fuser_basic.h
sergeisakov: Replaced `ConstRefToGate` with `GateToConstRef`.

sergeisakov: Replaced `ConstRefToGate` with `GateToConstRef`.

Issue #41, opened 2020-03-11
QSimSimulator: support non-zero initial states


95-martin-orion:
As highlighted in #40, the QSimSimulator only supports an all-zero initial state. To fix this, we should inject the necessary X gates at the start of the circuit when a non-zero initial state is requested.

sergeisakov:
In principle, qsim supports non-zero initial states. Injecting the necessary X gates might lead to unnecessary performance overhead. It should be done in a different way.

rmlarose:
I was going to raise an issue on unexpected behavior when providing initial states, then I found this one. Until a longer-term solution emerges, do you think it would be worthwhile to raise a warning/error when `initial_state` is not `None`? (This certainly would have saved me some debugging time!)

95-martin-orion:
That's a good point, @rmlarose - I'll add an error for this.

@sergeisakov, what do you have in mind for this? As it stands users need to manually add the X gates anyways, so in lieu of a more efficient solution injecting X gates at least wouldn't make performance any worse than it already is.

sergeisakov:
I was wrong.  Adding the necessary X gates should not lead to any performance overhead as those X gates will be fused to some other gates. However, adding X gates will support only classical initial states. The support for general initial states requires some additions to qsimcirq (note that the qsim library supports general initial states) in `pybind_interface/pybind_main.cpp` and in `qsimcirq/qsim_simulator.py`.

PR #257, opened 2020-12-16
Raise error for (unsupported) initial_state.

95-martin-orion:
In lieu of initial state support in qsimcirq, return an error when users request an initial state.


sergeisakov (approved):

95-martin-orion:
Related: #41

PR #258, opened 2020-12-17
Add quantum trajectory simulator.

sergeisakov:


95-martin-orion:
I'm taking vacation for most of the next two weeks, but I'll be in next Wednesday and should be able to review this then.

tests/qtrajectory_test.cc
```diff
@@ -0,0 +1,439 @@
...
+
+namespace {
+
+namespace types {
+
+using StateSpace = Simulator<For>::StateSpace;
+using State = StateSpace::State;
+using fp_type = StateSpace::fp_type;
+using Gate = Cirq::GateCirq<fp_type>;
+using QTSimulaltor = QuantumTrajectorySimulator<IO, Gate, MultiQubitGateFuser,
```
95-martin-orion, line 39: `QTSimulaltor` -> `QTSimulator`

tests/qtrajectory_test.cc
```diff
@@ -0,0 +1,439 @@
...
+       {normal, 0, p3, {M::Create(time, 0, {0, 0, 0, 0, s2, 0, 0, 0})}}});
+  ncircuit.push_back(
+      {{normal, 0, p1, {M::Create(time, 1, {t1, 0, 0, 0, 0, 0, r1, 0})}},
+       {normal, 0, p2, {M::Create(time, 1, {r2, 0, 0, 0, 0, 0, t2, 0})}},
+       {normal, 0, p3, {M::Create(time, 1, {0, 0, s1, 0, 0, 0, 0, 0})}},
+       {normal, 0, p3, {M::Create(time, 1, {0, 0, 0, 0, s2, 0, 0, 0})}}});
+}
+
+template <typename AddNoise>
+types::NoisyCircuit GenerateNoisyCirquit(double p, AddNoise&& add_noise) {
```
95-martin-orion, line 136: `GenerateNoisyCirquit` -> `GenerateNoisyCircuit`?

tests/qtrajectory_test.cc
```diff
@@ -0,0 +1,439 @@
...
+  ncircuit.push_back({{normal, 1, 1.0, {IS::Create(10, 0, 1)}}});
+  add_noise(11, p, ncircuit);
+  ncircuit.push_back({{KrausOperator<types::Gate>::kMeasurement, 1, 1.0,
+                       {gate::Measurement<types::Gate>::Create(12, {0, 1})}}});
+  add_noise(13, p, ncircuit);
+
+  return ncircuit;
+}
+
+void Run1(const types::NoisyCircuit& ncircuit,
```
95-martin-orion, line 171: Is the only distinction between `Run1` and `Run2` that `Run1` calls the run-multiple method once, while `Run2` calls the run-once method multiple times? If so, it might be useful to name these `RunBatch` and `RunOnceRepeatedly` (or something along those lines).

tests/qtrajectory_test.cc
```diff
@@ -0,0 +1,439 @@
...
+  std::vector<unsigned> histogram(1 << num_qubits, 0);
+
+  types::QTSimulaltor::Parameter param;
+  param.collect_mea_stat = true;
+
+  EXPECT_TRUE(types::QTSimulaltor::Run(param, num_qubits, ncircuit,
+                                       0, num_reps, measure, histogram));
+
+  for (std::size_t i = 0; i < histogram.size(); ++i) {
+    EXPECT_NEAR(double(histogram[i]) / num_reps, expected_results[i], 0.005);
```
95-martin-orion, line 192: Requiring accuracy within 0.5% sounds good to me - do you have a sense for how close we are to this boundary? Using fixed seeds should prevent flakes, but it'd be nice to know whether future changes to the operator-selection code are at risk of triggering a failure here.

tests/qtrajectory_test.cc
```diff
@@ -0,0 +1,439 @@
...
+  }
+}
+
+}  // namespace
+
+TEST(QTrajectoryTest, BitFlip) {
+/* The expected results are obtained with the following Cirq code.
```
95-martin-orion, line 232: [No change required] Thanks for these code snippets - it will make it much easier to construct the Cirq-to-qsim translation layer :)

tests/qtrajectory_test.cc
```diff
@@ -0,0 +1,439 @@
...
+                    const std::vector<uint64_t>& stat,
+                    std::vector<unsigned>& histogram) {
+    ASSERT_EQ(stat.size(), 1);
+    ++histogram[stat[0]];
+  };
+
+  std::vector<unsigned> histogram(1 << num_qubits, 0);
+
+  types::QTSimulaltor::Parameter param;
+  param.collect_mea_stat = true;
```
95-martin-orion, line 186: It looks like these tests don't cover the `collect_kop_stat = true` case. I'm fine with that for now given that Cirq also lacks that functionality, but could you add a TODO to test this once it's available in Cirq?

lib/qtrajectory.h
```diff
@@ -0,0 +1,435 @@
...
+   */
+  bool unitary;
+
+  /**
+   * Lower bound on Kraus operator probability.
+   */
+  double prob;
+
+  /**
+   * Sequence of operations that represet the Kraus operator. This can be just
```
95-martin-orion, line 54: `represet` -> `represent`

lib/qtrajectory.h
```diff
@@ -0,0 +1,435 @@
...
+using Channel = std::vector<KrausOperator<Gate>>;
+
+/**
+ * Noisy circuit.
+ */
+template <typename Gate>
+using NoisyCircuit = std::vector<Channel<Gate>>;
+
+/**
+ * Qauntum trajectory simulator.
```
95-martin-orion, line 73: `Qauntum` -> `Quantum`

tests/qtrajectory_test.cc
```diff
@@ -0,0 +1,439 @@
...
+  std::vector<double> expected_results = {
+    0.318501, 0.260538, 0.164616, 0.256345,
+  };
+
+  auto ncircuit = GenerateNoisyCirquit(0.1, AddGenAmplDumpNoise);
+  Run2(ncircuit, expected_results);
+}
+
+
+}  // namespace qsim
```
95-martin-orion, line 434: It would be nice to also have a test case verifying that the final state produced by the trajectory simulator is correct, or to update `Run1` and `Run2` to perform such a check.

95-martin-orion (changes requested):
I'm excited to see this feature coming to fruition! My first impression is that translation from Cirq `Channel` and `Mixture` objects to the qsim `Channel` type should be pretty simple, but we'll need Cirq support for Kraus index sampling before all of the features are available to Cirq users.

sergeisakov:
Thank you for the review.

lib/qtrajectory.h
```diff
@@ -0,0 +1,435 @@
...
+   *   (const State&), statistics of sampled Kraus operator indices and/or
+   *   measured bitstrings (const std::vector<uint64_t>&)] and any number of
+   *   optional parameters.
+   * @param args Optional arguments for the 'measure' function.
+   * @return True if the simulation completed successfully; false otherwise.
+   */
+  template <typename MeasurementFunc, typename... Args>
+  static bool Run(const Parameter& param, unsigned num_qubits,
+                  const NoisyCircuit<Gate>& circuit, uint64_t r0, uint64_t r1,
+                  MeasurementFunc&& measure, Args&&... args) {
```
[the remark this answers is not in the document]
sergeisakov: `measure` is a callback function to access the resulting state for each iteration. If `measure` is a no-op then this `Run` function has no effect. If the user is interested only in the final state then second set of `Run` functions can be used.

lib/qtrajectory.h
```diff
@@ -0,0 +1,435 @@
...
+   * seeded by repetition ID.
+   * @param param Options for the quantum trajectory simulator.
+   * @param num_qubits The number of qubits acted on by 'circuit'.
+   * @param circuit The noisy circuit to be simulated.
+   * @param r0, r1 The range of repetition IDs [r0, r1) to perform repetitions.
+   * @param measure Function that performs measurements (in the sense of
+   *   computing expectation values, etc). This function should have three
+   *   required parameters [repetition ID (uint64_t), final state vector
+   *   (const State&), statistics of sampled Kraus operator indices and/or
+   *   measured bitstrings (const std::vector<uint64_t>&)] and any number of
```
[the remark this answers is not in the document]
sergeisakov: Yes, this vector captures each Kraus index / measurement value in the order they appear in the circuit. It can contain a mix of both.

lib/qtrajectory.h
```diff
@@ -0,0 +1,435 @@
...
+
+        if (r < cp) {
+          DeferOps(kop.ops, gates);
+          CollectStat(param.collect_kop_stat, i, stat);
+
+          unitary = unitary && kop.unitary;
+
+          break;
+        }
+      }
```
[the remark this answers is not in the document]
sergeisakov: Yes, that is correct.

tests/qtrajectory_test.cc
sergeisakov: Done.

tests/qtrajectory_test.cc
sergeisakov: Done.

tests/qtrajectory_test.cc
sergeisakov: Done.

tests/qtrajectory_test.cc
sergeisakov: We are close to this boundary. The largest difference is about 0.00465 for 20000 repetitions. The accuracy depends on the number of repetitions and it is of course much better for, say, 10000000 repetitions. However, this leads to much longer test run times. Yes, future changes to the operator-selection code are at some risk of triggering a failure here. This can also be fixed by increasing the number of repetitions.

tests/qtrajectory_test.cc
sergeisakov: Good.

tests/qtrajectory_test.cc
sergeisakov: Added such a test.

lib/qtrajectory.h
sergeisakov: Done.

lib/qtrajectory.h
sergeisakov: Done.

tests/qtrajectory_test.cc
sergeisakov: What do you mean by "correct"? The final state after each iteration is the result of some stochastic process and it is random. I guess one way to check for correctness is to run a clean circuit and to compare with the clean circuit simulator. I added such a test.

95-martin-orion: You make a good point - I think the clean-circuit test covers the concerns I had in the original comment.

95-martin-orion (approved):
This looks ready to merge to me! One comment on the qsimcirq integration: we're still working out the API for expectation values in Cirq, so there may be some delay before I'm able to connect this to the Cirq side.

Issue #263, opened 2021-01-12
Multi qubit gate support for unitary


MichaelBroughton:
In order to have full support for multi qubit control gates in tensorflow quantum, we need to be able to use multi qubit control gates in the unitary calculator: https://github.com/quantumlib/qsim/blob/master/lib/unitary_calculator_basic.h#L112 . Which is blocking us on: https://github.com/tensorflow/quantum/issues/422 .  It looks like now all of the gate application logic in `simulator_***.cc` is generated by code that writes these C++ functions now so it would be pretty hard for a human being to implement this in a consistent fashion without the scripts mentioned here #227 . @sergeisakov would you be able to implement the logic for multi qubit controlled gate application for the unitary calculator ?

sergeisakov:
Yes, all of the gate application logic in `simulator_***.cc` is generated by a script. However, it doesn't have to be generated by a script. A generic version can be provided (with some performance loss for AVX* and perhaps for SSE) or function templates can be used (without performance loss). Code generation really makes sense only for AVX and AVX512 versions of the simulator. The unitary calculator has only a basic implementation and a generic multi-qubit gate application logic can be written by a human being without a script. I think I'm not the right person to do that.

MichaelBroughton:
> I think I'm not the right person to do that.

So as somebody who helps maintain this repo and is responsible for most of the major contributions by far, you think you aren't the person to do it :P ?

Either way, I can probably struggle my way through implementing this, if you really don't want to at any point. Would we prefer to do that instead ?

PR #264, opened 2021-01-19
Updated Unitary calculator following new code generation style.

MichaelBroughton:
Updates unitary calculator to use the same style of code in the rest of the library. Hopefully when #227 is resolved, support for unitary calculator can be included more long term :). This is the first step in resolving #263 .

### DIFF
--- a/lib/unitary_calculator_basic.h
+++ b/lib/unitary_calculator_basic.h
@@ -21,48 +21,7 @@
 #include "unitaryspace_basic.h"
 
 namespace qsim {
-
 namespace unitary {
-
-namespace {
-
-template <typename FP = float>
-inline const FP dot_one_r(const FP m_0i_r, const FP m_0i_i, const FP m_1i_r,
-                          const FP m_1i_i, const int i, const FP* gate) {
-  return m_0i_r * gate[i * 4 + 0] - m_0i_i * gate[i * 4 + 1] +
-         m_1i_r * gate[i * 4 + 2] - m_1i_i * gate[i * 4 + 3];
-}
-
-template <typename FP = float>
-inline const FP dot_one_i(const FP m_0i_r, const FP m_0i_i, const FP m_1i_r,
-                          const FP m_1i_i, const int i, const FP* gate) {
-  return m_0i_r * gate[i * 4 + 1] + m_0i_i * gate[i * 4 + 0] +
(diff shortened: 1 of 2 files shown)


lib/unitary_calculator_basic.h
```diff
@@ -197,209 +161,76 @@ class UnitaryCalculatorBasic final {
...
+                      unsigned sqrt_size, fp_type* rstate) {
+      fp_type rn, in;
+      fp_type rs[16], is[16];
+
+      auto row_size = uint64_t{1} << n_qb;
+      uint64_t i = ii % sqrt_size;
+      uint64_t j = ii / sqrt_size;
+
+      uint64_t k = (1 * i & ms[0]) | (2 * i & ms[1]) | (4 * i & ms[2]);
+      uint64_t kk = (1 * j & ms[0]) | (2 * j & ms[1]) | (4 * j & ms[2]);
```
95-martin-orion, line 197: Same here - rename to avoid collision with `k` used in loop.

95-martin-orion (approved):
Looks good modulo some variable names. Is my understanding correct that the changes between `simulator_basic` and this code are due to manipulating a vector vs. the unitary matrix?

lib/unitary_calculator_basic.h
```diff
@@ -121,68 +80,73 @@ class UnitaryCalculatorBasic final {
...
+      fp_type rn, in;
+      fp_type rs[4], is[4];
+
+      auto row_size = uint64_t{1} << n_qb;
+
+      uint64_t i = ii % sqrt_size;
+      uint64_t j = ii / sqrt_size;
+
+      uint64_t k = (1 * i & ms[0]) | (2 * i & ms[1]);
+      uint64_t kk = (1 * j & ms[0]) | (2 * j & ms[1]);
```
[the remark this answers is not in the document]
MichaelBroughton: Fixed.

lib/unitary_calculator_basic.h
MichaelBroughton: Done.

MichaelBroughton:
> Is my understanding correct that the changes between simulator_basic and this code are due to manipulating a vector vs. the unitary matrix?

Yes. The unitary calculator is just like the state vector simulators with two major changes:
1. Larger underlying array (grows like 2**n * 2**n instead of 2**n)
2. The indexing routines and gate application logic have to account for the row and column when applying the gate instead of just the one long row (or column) that is done in the state vector simulator.

PR #265, opened 2021-01-20
Add single qubit gate control support.

MichaelBroughton:
Adds support for single qubit controlled gates in the unitary calculator. Tries to follow as closely as possible with the code generation style used in the rest of the library.


sergeisakov (approved):
Looks good to me.

PR #266, opened 2021-01-22
Added two qubit controlled gates to unitary.

### LINKED ISSUE #263: Multi qubit gate support for unitary

MichaelBroughton:
Adds support for two qubit multi control gates. Fixes #263 and unblocks https://github.com/tensorflow/quantum/pull/458 pending an upgrade to the latest qsim in TFQ.


sergeisakov (approved):
LGTM

PR #268, opened 2021-01-24
Bulk set

MichaelBroughton:
This would be a really nice to have feature to use in TFQ. Are we OK with adding a utility function to statespace that behaves the same as `SetAmpl`, but can do multiple assignments at once ?


sergeisakov (approved):
LGTM. Thank you for adding that. This is a useful feature.

PR #274, opened 2021-01-26
Add inclusion/exclusion support on bulksetampl.

MichaelBroughton:
Adds the `exclude` flag to BulkSetAmpl to provide greater flexibility for bulk writing. Would be helpful on https://github.com/tensorflow/quantum/pull/458

### DIFF
--- a/tests/statespace_testfixture.h
+++ b/tests/statespace_testfixture.h
@@ -820,7 +820,7 @@ void TestBulkSetAmplitude() {
   for(int i = 0; i < 8; i++) {
     state_space.SetAmpl(state, i, 1, 1);
   }
-  state_space.BulkSetAmpl(state, 1, 0, 0, 0);
+  state_space.BulkSetAmpl(state, 1, 0, 0, 0, false);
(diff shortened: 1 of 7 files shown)


sergeisakov (approved):

sergeisakov:
LGTM

tests/statespace_testfixture.h
```diff
@@ -820,7 +820,7 @@ void TestBulkSetAmplitude() {
   for(int i = 0; i < 8; i++) {
     state_space.SetAmpl(state, i, 1, 1);
   }
-  state_space.BulkSetAmpl(state, 1, 0, 0, 0);
+  state_space.BulkSetAmpl(state, 1, 0, 0, 0, false);
```
95-martin-orion, line 823: Since `false` is the default, we should have at least one test case where this argument is omitted.

95-martin-orion (changes requested):
One nit on the tests, but the rest looks alright to me.

tests/statespace_testfixture.h
MichaelBroughton: Done.

95-martin-orion (approved):

PR #275, opened 2021-01-27
Fix _resolve_parameters_ args.

95-martin-orion:
Updates parameter resolution in qsimcirq to match Cirq. Related: https://github.com/quantumlib/Cirq/pull/3546


95-martin-orion:
Jumped on this too soon! Need to wait for Cirq v0.10.

PR #280, opened 2021-01-29
Update unitary calculator.

sergeisakov:
Gate application is simplified. This code is generated.

### DIFF
--- a/lib/unitary_calculator_basic.h
+++ b/lib/unitary_calculator_basic.h
@@ -15,8 +15,10 @@
 #ifndef UNITARY_CALCULATOR_BASIC_H_
 #define UNITARY_CALCULATOR_BASIC_H_
 
+
+#include <algorithm>
+#include <complex>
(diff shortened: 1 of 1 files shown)


lib/unitary_calculator_basic.h
```diff
@@ -15,8 +15,10 @@
 #ifndef UNITARY_CALCULATOR_BASIC_H_
 #define UNITARY_CALCULATOR_BASIC_H_
 
+
+#include <algorithm>
+#include <complex>
```
MichaelBroughton, line 20: Why add these ?

MichaelBroughton (approved):
LGTM

lib/unitary_calculator_basic.h
sergeisakov: Thanks, that's a good point. I'll remove that in the next PR when I add support for multi-qubit gates.

PR #281, opened 2021-02-03
Add threading test and increase coverage.

MichaelBroughton:


95-martin-orion (approved):
Tests are passing, and the necessary import transforms are fairly simple - LGTM.

I'll need to get those transforms submitted before my next import attempt, but that's not a blocker here.

tests/simulator_avx_test.cc
```diff
@@ -16,45 +16,58 @@
 
 #include "gtest/gtest.h"
 
-#include "../lib/formux.h"
+#include "../lib/parfor.h"
+#include "../lib/seqfor.h"
```
95-martin-orion, line 20: From the Kokoro failures, I would guess that these need to be in the `ifdef` block, similar to `formux.h`.

PR #282, opened 2021-02-04
Don't overwrite initial states in qtrajectory

95-martin-orion:
Came across this in testing for qsimcirq integration: if we `SetStateZero` for all states, users can't specify an input state.


95-martin-orion:
It looks like this isn't the correct solution here, but the issue stands. Opened #283 to track.

PR #284, opened 2021-02-04
Don't overwrite initial states in qtrajectory

### LINKED ISSUE #283: qtrajectory overwrites initial states
Came across this in testing for qsimcirq integration: input states for qtrajectory get overwritten.
https://github.com/quantumlib/qsim/blob/6ec7e8ded104b984e49f743dd8c5f7566276ff92/lib/qtrajectory.h#L239-L246

Tried the obvious solution in #282, but that broke several tests.

sergeisakov:
Closes #283

### DIFF
--- a/lib/qtrajectory.h
+++ b/lib/qtrajectory.h
@@ -154,12 +154,17 @@ class QuantumTrajectorySimulator {
     std::vector<const Gate*> gates;
     gates.reserve(4 * std::size_t(clast - cfirst));
 
-    State state = StateSpace(1).Null();
-    State scratch = StateSpace(1).Null();
+    StateSpace state_space(param.num_threads);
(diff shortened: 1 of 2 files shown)


95-martin-orion (changes requested):
Could you also add a test to verify this? It'll be covered by the eventual qsimcirq tests, but I think it's worthwhile to confirm it at this level as well.

Thanks for the quick fix!

sergeisakov:
> Could you also add a test to verify this works with initial states? It'll be covered by the eventual qsimcirq tests, but I think it's worthwhile to confirm it at this level as well.
> 
> Thanks for the quick fix!

Added a test.

95-martin-orion (approved):

PR #286, opened 2021-02-09
Don't swap states in qtrajectory.

sergeisakov:


tests/qtrajectory_test.cc
```diff
@@ -217,6 +219,8 @@ void RunOnceRepeatedly(const types::NoisyCircuit& ncircuit,
     EXPECT_TRUE(types::QTSimulator::Run(param, num_qubits, ncircuit, i,
                                         scratch, state, stat));
 
+    EXPECT_EQ(state_pointer, state.get());
```
95-martin-orion, line 222: As good a test as any, I suppose :)

95-martin-orion (approved):
LGTM - looks like this uses the same number of copy-operations, just at a different location in the code.

sergeisakov:
> LGTM - looks like this uses the same number of copy-operations, just at a different location in the code.

Thank you for the review. Yes, this uses the same number of copy-operations.

PR #287, opened 2021-02-09
Enable trajectory simulation via qsimcirq

### LINKED ISSUE #270: Expose trajectory simulation methods through qsimcirq
Trajectory simulation is currently available in qsim-core, but inaccessible to Cirq users. The qsimcirq translation layer must be updated to accept noisy circuits (i.e. circuits with Channels and Mixtures) and route them to the proper qsim-core methods.

95-martin-orion:
Fixes #270. Since this is a fairly large PR, below is a high-level overview of what's changing.

-----
`qtrajectory.h`:
* Added a `qubit_count` method to calculate `num_qubits` from a qsim `Circuit`

`pybind_main.(h|cpp)`:
* Split the `add_gate` methods into `create_gate` and `add_gate` methods
* Added `add_gate_channel` methods for adding single-gate (i.e. noiseless) channels to a `NoisyCircuit`
* Added `add_channel` for adding Cirq `Mixture` and `Channel` objects to a `NoisyCircuit`
* Added qtrajectory equivalents of all `qsim_(simulate|sample)` methods

`qsim_circuit.py`:
* Split "add operation to circuit" behavior into a helper method
* Added `translate_cirq_to_qtrajectory`, which translates arbitrary noisy Cirq circuits to qsim `NoisyCircuit`s

`qsim_simulator.py`:
* Added `_needs_trajectories` to check if a circuit has noise (and thus requires trajectory simulation)
    * This should be added to Cirq, but until the next release we won't have access to it.
* Updated `simulate` and `sample` methods to call qtrajectory methods when needed

`qsimcirq_test.py`:
* Parameterized existing `simulate` and `sample` tests to run in both qsim and qtrajectory modes
* Added tests for noisy circuit simulation

### DIFF
--- a/lib/qtrajectory.h
+++ b/lib/qtrajectory.h
@@ -68,6 +68,21 @@ using Channel = std::vector<KrausOperator<Gate>>;
 template <typename Gate>
 using NoisyCircuit = std::vector<Channel<Gate>>;
 
+template <typename Gate>
+unsigned count_qubits(const NoisyCircuit<Gate>& ncircuit) {
+  std::set<int> qubit_set;
+  for (const Channel<Gate>& channel : ncircuit) {
+    for (const KrausOperator<Gate>& kop : channel) {
+      for (const Gate& gate : kop.ops) {
+        for (const unsigned& qidx : gate.qubits) {
+          qubit_set.insert(qidx);
+        }
+      }
+    }
+  }
+  return qubit_set.size();
+}
+
 /**
  * Quantum trajectory simulator.
  */
--- a/lib/qtrajectory.h
+++ b/lib/qtrajectory.h
@@ -68,21 +68,6 @@ using Channel = std::vector<KrausOperator<Gate>>;
 template <typename Gate>
 using NoisyCircuit = std::vector<Channel<Gate>>;
 
-template <typename Gate>
-unsigned count_qubits(const NoisyCircuit<Gate>& ncircuit) {
-  std::set<int> qubit_set;
-  for (const Channel<Gate>& channel : ncircuit) {
-    for (const KrausOperator<Gate>& kop : channel) {
-      for (const Gate& gate : kop.ops) {
-        for (const unsigned& qidx : gate.qubits) {
-          qubit_set.insert(qidx);
-        }
-      }
-    }
-  }
-  return qubit_set.size();
-}
-
 /**
  * Quantum trajectory simulator.
  */
--- a/pybind_interface/pybind_main.cpp
+++ b/pybind_interface/pybind_main.cpp
@@ -55,6 +55,14 @@ Circuit<Cirq::GateCirq<float>> getCircuit(const py::dict &options) {
   }
 }
 
+NoisyCircuit<Cirq::GateCirq<float>> getNoisyCircuit(const py::dict &options) {
+  try {
+    return options["c\0"].cast<NoisyCircuit<Cirq::GateCirq<float>>>();
+  } catch (const std::invalid_argument &exp) {
+    throw;
+  }
+}
+
 std::vector<Bitstring> getBitstrings(const py::dict &options, int num_qubits) {
   std::string bitstrings_str;
   try {
@@ -307,10 +315,6 @@ void control_last_gate_channel(const std::vector<unsigned>& qubits,
   }
 }
 
-int count_gates(NoisyCircuit<Cirq::GateCirq<float>> ncircuit) {
-  return ncircuit.size();
-}
-
 // TODO: need methods for creating Kraus ops and channels
 
 std::vector<std::complex<float>> qsim_simulate(const py::dict &options) {
@@ -358,8 +362,57 @@ std::vector<std::complex<float>> qsim_simulate(const py::dict &options) {
 }
 
 std::vector<std::complex<float>> qtrajectory_simulate(const py::dict &options) {
-  // TODO: implement
-  return {};
+  NoisyCircuit<Cirq::GateCirq<float>> ncircuit;
+  unsigned num_qubits;
+  std::vector<Bitstring> bitstrings;
+  try {
+    ncircuit = getNoisyCircuit(options);
+    num_qubits = qsim::count_qubits(ncircuit);
+    bitstrings = getBitstrings(options, num_qubits);
+  } catch (const std::invalid_argument &exp) {
+    IO::errorf(exp.what());
+    return {};
+  }
+
+  using Simulator = qsim::Simulator<For>;
+  using StateSpace = Simulator::StateSpace;
+  using State = StateSpace::State;
+
+  // Define container for amplitudes
+  std::vector<std::complex<float>> amplitudes;
+  amplitudes.reserve(bitstrings.size());
+
+  using Runner = qsim::QuantumTrajectorySimulator<IO, Cirq::GateCirq<float>,
+                                                  MultiQubitGateFuser,
+                                                  Simulator>;
+
+  Runner::Parameter param;
+  try {
+    param.num_threads = parseOptions<unsigned>(options, "t\0");
+    param.max_fused_size = parseOptions<unsigned>(options, "f\0");
+    param.verbosity = parseOptions<unsigned>(options, "v\0");
+    // TODO: check with Sergei (move from method argument?)
+    // param.seed = parseOptions<unsigned>(options, "s\0");
+  } catch (const std::invalid_argument &exp) {
+    IO::errorf(exp.what());
+    return {};
+  }
+
+  StateSpace state_space(param.num_threads);
+  auto measure = [&bitstrings, &ncircuit, &amplitudes, &state_space](
+                  unsigned k, const State &state,
+                  std::vector<uint64_t>& stat) {
+    for (const auto &b : bitstrings) {
+      amplitudes.push_back(state_space.GetAmpl(state, b));
+    }
+  };
+
+  uint64_t seed = 1;
+  if (!Runner::Run(param, num_qubits, ncircuit, seed, seed + 1, measure)) {
+    IO::errorf("qtrajectory simulation of the circuit errored out.\n");
+    return {};
+  }
+  return amplitudes;
 }
 
 // Simulate from a "pure" starting state.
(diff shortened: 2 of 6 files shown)


pybind_interface/pybind_main.cpp
```diff
@@ -73,264 +82,269 @@ std::vector<Bitstring> getBitstrings(const py::dict &options, int num_qubits) {
...
+  //   0. The lower-bound probability of applying the matrix.
+  //   1. The matrix to be applied.
+  //   2. Whether the matrix is unitary.
+  for (const auto &triple : prob_matrix_unitary_triples) {
+    const float prob = std::get<0>(triple);
+    const std::vector<float>& mat = std::get<1>(triple);
+    bool is_unitary = std::get<2>(triple);
+    Gate gate = create_matrix_gate(time, qubits, mat);
+    channel.emplace_back(KrausOperator<Gate>{
+      KrausOperator<Gate>::kNormal, is_unitary, prob, {gate}
```
MichaelBroughton, line 340: Would the need for kMeasurement never come up ?

pybind_interface/pybind_main.cpp
```diff
@@ -375,6 +389,59 @@ std::vector<std::complex<float>> qsim_simulate(const py::dict &options) {
...
+    IO::errorf(exp.what());
+    return {};
+  }
+
+  StateSpace state_space(param.num_threads);
+  auto measure = [&bitstrings, &ncircuit, &amplitudes, &state_space](
+                  unsigned k, const State &state,
+                  std::vector<uint64_t>& stat) {
+    for (const auto &b : bitstrings) {
+      amplitudes.push_back(state_space.GetAmpl(state, b));
```
MichaelBroughton, line 434: Based on the name of this function `qtrajectory_simulate` why are you keep track of the raw amplitudes you get for each bitstring you draw ?

pybind_interface/pybind_main.cpp
```diff
@@ -375,6 +389,59 @@ std::vector<std::complex<float>> qsim_simulate(const py::dict &options) {
...
+    return {};
+  }
+
+  using Simulator = qsim::Simulator<For>;
+  using StateSpace = Simulator::StateSpace;
+  using State = StateSpace::State;
+
+  // Define container for amplitudes
+  std::vector<std::complex<float>> amplitudes;
+  amplitudes.reserve(bitstrings.size());
```
MichaelBroughton, line 411: Is this reserve always for size zero ?

pybind_interface/pybind_main.cpp
```diff
@@ -54,6 +55,14 @@ Circuit<Cirq::GateCirq<float>> getCircuit(const py::dict &options) {
   }
 }
 
+NoisyCircuit<Cirq::GateCirq<float>> getNoisyCircuit(const py::dict &options) {
+  try {
+    return options["c\0"].cast<NoisyCircuit<Cirq::GateCirq<float>>>();
```
MichaelBroughton, line 60: Is the compiler smart enough to turn this into a move ? I would assume yes, but am not 100% sure.

pybind_interface/pybind_main.cpp
```diff
@@ -489,6 +556,132 @@ py::array_t<float> qsim_simulate_fullstate(
...
+    IO::errorf(exp.what());
+    return {};
+  }
+
+  StateSpace state_space(param.num_threads);
+
+  float *fsv;
+  const uint64_t fsv_size = std::pow(2, num_qubits + 1);
+  const uint64_t buff_size = state_space.MinSize(num_qubits);
+  if (posix_memalign((void **)&fsv, 32, buff_size * sizeof(float))) {
```
MichaelBroughton, line 595: This needs posix platforms and might not work on things like windows. See https://github.com/quantumlib/qsim/blob/master/lib/vectorspace.h#L74

pybind_interface/pybind_main.cpp
```diff
@@ -489,6 +556,132 @@ py::array_t<float> qsim_simulate_fullstate(
...
+    IO::errorf("Memory allocation failed.\n");
+    return {};
+  }
+
+  State state = state_space.Create(fsv, num_qubits);
```
MichaelBroughton, line 600: Related to above question: why use this creation method instead of the default  Create() ?

pybind_interface/pybind_main.cpp
```diff
@@ -542,6 +735,75 @@ std::vector<unsigned> qsim_sample(const py::dict &options) {
...
+    param.max_fused_size = parseOptions<unsigned>(options, "f\0");
+    param.verbosity = parseOptions<unsigned>(options, "v\0");
+    seed = parseOptions<unsigned>(options, "s\0");
+    param.collect_mea_stat = true;
+  } catch (const std::invalid_argument &exp) {
+    IO::errorf(exp.what());
+    return {};
+  }
+
+  std::vector<std::vector<unsigned>> results;
```
MichaelBroughton, line 769: Would it be worth reserving sizes here ?

qsimcirq/qsim_circuit.py
```diff
@@ -212,45 +276,100 @@ def to_matrix(op: cirq.ops.GateOperation):
...
+            square_mat = np.reshape(mat, (int(np.sqrt(mat.size)), -1))
+            unitary = cirq.is_unitary(square_mat)
+            singular_vals = np.linalg.svd(square_mat)[1]
+            lower_bound_prob = min(singular_vals) ** 2
+            # Package matrix into a qsim-friendly format.
+            mat = np.reshape(mat, (-1,)).astype(np.complex64, copy=False)
+            chdata.append((lower_bound_prob, mat.view(np.float32), unitary))
+          qubits = [qubit_to_index_dict[q] for q in channel.qubits]
+          qsim.add_channel(time_offset, qubits, chdata, qsim_ncircuit)
+      time_offset += moment_length
```
MichaelBroughton, line 373: It seems like doing this math in python and then passing these matrices through the translation might incur a lot of uneseccesary overhead vs just doing this SVD on the qsim side. Maybe this would be a good time to add Eigen as a dependency ?

MichaelBroughton (commented):
Looks good. Kind of hard to review so much code and catch everything that is going on. Might be worth breaking down. The only thing that really caught my eye as a little strange was the use of `posix_memalign` in the pybind code along with the particular version of `Create()` being used.

pybind_interface/pybind_main.cpp
95-martin-orion: It's not needed in `add_channel`, since measurements don't appear in channels. A measurement gate would instead be handled by `create_single_gate_channel`.

pybind_interface/pybind_main.cpp
95-martin-orion: This follows the previous naming pattern of `*_simulate` for amplitudes and `*_simulate_fullstate` for state vector. As noted in the PR description, each `qtrajectory_{foo}` method is the noisy equivalent of the `qsim_{foo}` method.

pybind_interface/pybind_main.cpp
95-martin-orion: No - `bitstrings` contains the list of bitstrings for which the user has requested amplitudes. This reserves an equal number of spaces for storing amplitudes.

pybind_interface/pybind_main.cpp
95-martin-orion: Not sure. This is a copy of the `getCircuit` method, though, so it should follow our pre-existing behavior.

pybind_interface/pybind_main.cpp
95-martin-orion: This is copied from the analogous `qsim_simulate_fullstate` method. + @sergeisakov for insight on copied code.

pybind_interface/pybind_main.cpp
95-martin-orion: I don't think we have cheap access to # of measurements here. Unles we just want to reserve an arbitrary, large amount of space, it's probably not worth it.

qsimcirq/qsim_circuit.py
95-martin-orion: I won't say this is infeasible, but it would add a degree of complexity - both to this PR and to the repository as a whole, since we would need to add dependency on Eigen (likely similar to the existing googletest submodule).

How much overhead do you see this adding? I wasn't able to find a concrete answer on Eigen vs. numpy performance in a quick search, but with most operators being 1- or 2-qubit I wouldn't expect there to be much difference.

95-martin-orion: On a related note: during design we considered putting SVD calculation in C++ so that our C++ users could access it, but IIRC we decided against it under the assumption that C++ users would prefer precalculating their lower bounds for improved performance.

sergeisakov: I think this shouldn't be a significant overhead, especially, when this is done once for a large number of trajectories.

sergeisakov (changes requested):
Great work, thank!

qsimcirq/qsim_circuit.py
```diff
@@ -143,6 +144,69 @@ def _control_details(gate: cirq.ops.ControlledGate, qubits):
...
+      qsim.add_diagonal_gate if isinstance(circuit, qsim.Circuit)
+      else qsim.add_diagonal_gate_channel
+    )
+    op(time, qsim_qubits, qsim_gate._diag_angles_radians, circuit)
+  elif gate_kind == qsim.kMatrixGate:
+    op = (
+      qsim.add_matrix_gate if isinstance(circuit, qsim.Circuit)
+      else qsim.add_matrix_gate_channel
+    )
+    m = [
```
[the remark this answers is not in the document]
95-martin-orion: From what I can find, [there is no builtin `flatten` method for Python](https://softwareengineering.stackexchange.com/questions/254279/why-doesnt-python-have-a-flatten-function-for-lists).

lib/qtrajectory.h
```diff
@@ -68,6 +68,21 @@ using Channel = std::vector<KrausOperator<Gate>>;
 template <typename Gate>
 using NoisyCircuit = std::vector<Channel<Gate>>;
 
+template <typename Gate>
+unsigned count_qubits(const NoisyCircuit<Gate>& ncircuit) {
```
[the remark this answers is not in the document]
95-martin-orion: +1 to a struct for NoisyCircuit. I think that would also clean up some of the pybind code, since it's a bit finicky about binding STL types to Python objects.

Any preference on the number-of-qubits option name? I hesitate to use `-q` in a library where so many things start with Q - would the current code support a `--num_qubits` option?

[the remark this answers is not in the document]
sergeisakov: I would use `-n` for  the number-of-qubits option name.

pybind_interface/pybind_main.cpp
95-martin-orion: Opened #289 to track this. If we get a fix for that before this goes in, I'll update this PR to match - otherwise leaving as-is.

qsimcirq/qsim_circuit.py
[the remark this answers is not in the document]
sergeisakov: Yes, you are right.

qsimcirq/qsim_circuit.py
```diff
@@ -143,6 +144,69 @@ def _control_details(gate: cirq.ops.ControlledGate, qubits):
...
+    qsim_qubits = qubits[qsim_gate.num_controls():]
+    qsim_gate = qsim_gate.sub_gate
+
+  if (gate_kind == qsim.kTwoQubitDiagonalGate or
+      gate_kind == qsim.kThreeQubitDiagonalGate):
+    op = (
+      qsim.add_diagonal_gate if isinstance(circuit, qsim.Circuit)
+      else qsim.add_diagonal_gate_channel
+    )
+    op(time, qsim_qubits, qsim_gate._diag_angles_radians, circuit)
```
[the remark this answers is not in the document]
95-martin-orion: There isn't a good one - I think I just wrote this right after the `if(_needs_trajectories)` code, and was stuck in a functions-as-variables mindset. Will switch to the format you describe.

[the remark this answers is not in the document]
95-martin-orion: Done.

lib/qtrajectory.h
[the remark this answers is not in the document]
95-martin-orion: Done.

pybind_interface/pybind_main.cpp
```diff
@@ -375,6 +389,59 @@ std::vector<std::complex<float>> qsim_simulate(const py::dict &options) {
   return amplitudes;
 }
 
+std::vector<std::complex<float>> qtrajectory_simulate(const py::dict &options) {
+  NoisyCircuit<Cirq::GateCirq<float>> ncircuit;
+  unsigned num_qubits;
+  std::vector<Bitstring> bitstrings;
+  try {
+    ncircuit = getNoisyCircuit(options);
+    num_qubits = qsim::count_qubits(ncircuit);
```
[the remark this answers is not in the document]

lib/qtrajectory.h
```diff
@@ -68,6 +68,21 @@ using Channel = std::vector<KrausOperator<Gate>>;
...
+  std::set<int> qubit_set;
+  for (const Channel<Gate>& channel : ncircuit) {
+    for (const KrausOperator<Gate>& kop : channel) {
+      for (const Gate& gate : kop.ops) {
+        for (const unsigned& qidx : gate.qubits) {
```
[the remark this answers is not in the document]
95-martin-orion: Removed `count_qubits`.

pybind_interface/pybind_main.cpp
95-martin-orion: Is there a change requested here? Otherwise I'll mark this resolved.

sergeisakov: I think no change is requested here. It might be worth addressing this issue in a later PR.

pybind_interface/pybind_main.cpp
sergeisakov: #290 can be used to fix it. Replace
```C++
  float *fsv;
  const uint64_t fsv_size = std::pow(2, num_qubits + 1);
  const uint64_t buff_size = state_space.MinSize(num_qubits);
  if (posix_memalign((void **)&fsv, 32, buff_size * sizeof(float))) {
    IO::errorf("Memory allocation failed.\n");
    return {};
  }

  State state = state_space.Create(fsv, num_qubits);
```
with
```C++
  State state = state_space.Create(num_qubits);
```
and add `float* fsv = state.release();` after `state_space.InternalToNormalOrder(state);`. One thing is that `delete` is used to deallocate this memory. It seems it works with `posix_memalign`. I don't know if it works on Windows. Perhaps `delete reinterpret_cast<float *>(data);` can be replaced with `detail::free(data);`.

95-martin-orion:
Cleaned up some extra `num_qubits` and `all_qubits` in the Python code to improve performance.

sergeisakov (approved):
LGTM

Issue #271, opened 2021-01-25
Implement Cirq expectation value API


95-martin-orion:
Cirq has introduced an API for requesting expectation values from a simulator: https://github.com/quantumlib/Cirq/issues/3492

We need to implement this interface in qsimcirq. This should be mostly orthogonal to #270, but all combinations should be supported in the end (noiseless full-state, noisy full-state, noiseless EVs, noisy EVs).

PR #291, opened 2021-02-22
Separate simulation and python conversion

95-martin-orion:
This is a cleanup step on the path to #271: the expectation value API will need the simulation half of this change, but will run `ExpectationValue` on the results.

I'm not 100% confident in my use of RVO and rvalue arguments here. @sergeisakov, could you verify that this does not introduce any additional copies of the state?


sergeisakov (approved):
LGTM. There are no any additional copies of the state (the state is not copiable anyway).

PR #294, opened 2021-02-24
Relax constraint on gate time order.

sergeisakov:


lib/fuser.h
```diff
@@ -51,6 +52,87 @@ struct GateFused {
...
+      if (gate.kind == gate::kMeasurement) {
+        if (epochs.size() == 0 || epochs.back() < gate.time) {
+          if (!AddBoundary(gate.time, max_time, epochs)) {
+            epochs.resize(0);
+            return epochs;
+          }
+        }
+      }
+
+      if (last < times.size() && gate.time >= times[last]) {
```
95-martin-orion, line 103: Is this `if` block necessary? Seems like the `while` condition is identical.

lib/fuser_basic.h
```diff
@@ -31,19 +31,13 @@ namespace qsim {
  * User-defined controlled gates (controlled_by.size() > 0) and gates acting on
  * more than two qubits are not fused.
  * The template parameter Gate can be Gate type or a pointer to Gate type.
+ * This class is deprecated.
```
95-martin-orion, line 34: What should previous users of this class use instead - I assume MultiQubitGateFuser? (Please add this in the comment here)

lib/fuser_basic.h
```diff
@@ -57,15 +51,17 @@ struct BasicGateFuser final {
   };
 
   /**
-   * Stores ordered sets of gates, each acting on two qubits, that can be
-   * applied together. Note that gates fused with this method are not
+   * Stores sets of gates that can be applied together. Only one- and
+   * two-qubit gates are get fused. Gates fused with this method are not
```
95-martin-orion, line 55: "are get" -> "will get"

Same elsewhere in file.

tests/fuser_basic_test.cc
```diff
@@ -1104,6 +1108,336 @@ TEST(FuserBasicTest, ControlledGate) {
...
+    std::vector<Gate> circuit = {
+      GateCZ<float>::Create(0, 0, 1),
+      GateCZ<float>::Create(0, 2, 3),
+      GateCZ<float>::Create(2, 0, 3),
+      GateCZ<float>::Create(1, 1, 2),
+    };
+
+    std::vector<unsigned> time_boudary = {1};
+    auto fused_gates = Fuser::FuseGates(
+        param, num_qubits, circuit.begin(), circuit.end(), time_boudary);
```
95-martin-orion, line 1355: `time_boudary` -> `time_boundary`

Same elsewhere in file.

tests/fuser_mqubit_test.cc
```diff
@@ -644,6 +648,333 @@ TEST(FuserMultiQubitTest, SmallCircuits) {
...
+      CreateDummyGate(0, {4, 5}),
+      CreateDummyGate(1, {3, 4}),
+      CreateDummyGate(1, {5, 0}),
+    };
+
+    param.max_fused_size = 3;
+    auto fused_gates = Fuser::FuseGates(
+        param, num_qubits, circuit.begin(), circuit.end());
+
+    EXPECT_EQ(fused_gates.size(), 4);
```
95-martin-orion, line 724: Which gates fuse in this example? I guess the reason it's 4 instead of 3 is that fusing the last pair would form a loop - i.e. `(4,5) -> (0,1)-(1,2) -> (2,3)-(3,4) -> (5,0)`?

tests/fuser_mqubit_test.cc
```diff
@@ -644,6 +648,333 @@ TEST(FuserMultiQubitTest, SmallCircuits) {
...
+      CreateDummyGate(3, {3, 4}),
+      CreateDummyGate(2, {5, 6}),
+    };
+
+    {
+      param.max_fused_size = 3;
+      auto fused_gates = Fuser::FuseGates(
+          param, num_qubits, circuit.begin(), circuit.end());
+
+      EXPECT_TRUE(TestFusedGates(num_qubits, circuit, fused_gates));
```
95-martin-orion, line 783: Should this test include a `fused_gates.size()` check? Same for the one immediately below this.

tests/qtrajectory_test.cc
```diff
@@ -132,8 +226,9 @@ void AddGenAmplDumpNoise(
        {normal, 0, p3, {M::Create(time, 1, {0, 0, 0, 0, s2, 0, 0, 0})}}});
 }
 
-template <typename AddNoise>
-types::NoisyCircuit GenerateNoisyCircuit(double p, AddNoise&& add_noise) {
+template <typename AddNoise1, typename AddNoise2>
+types::NoisyCircuit GenerateNoisyCircuit(
```
95-martin-orion, line 230: My read is that the core behavior of this test isn't changing, you've just updated it to take advantage of the looser time-order restriction. Is that correct?

95-martin-orion (changes requested):
Test coverage looks very thorough - comments are mostly clarifications and formatting nits.

lib/fuser.h
sergeisakov: You are right, thanks. Fixed.

lib/fuser_basic.h
sergeisakov: Done.

lib/fuser_basic.h
sergeisakov: Done.

tests/fuser_basic_test.cc
sergeisakov: Done.

tests/fuser_mqubit_test.cc
sergeisakov: Exactly, the last pair would form a loop and we have to avoid that.

tests/fuser_mqubit_test.cc
sergeisakov: Done.

tests/qtrajectory_test.cc
sergeisakov: Yes, that is correct. This basically shows an alternative way of constructing Kraus operators.

95-martin-orion (approved):

PR #298, opened 2021-03-02
Add AVX and SSE unitary calculators.

sergeisakov:


MichaelBroughton (approved):
Looks good. Cool trick on the SmallCircuitTest to hit all of the different cases in the ApplyGate functions.

Issue #288, opened 2021-02-10
Ensure 1:1 parity with Cirq channels


MichaelBroughton:
In order to make life easier for qsimcirq translations and TFQ translations, would it be possible to implement kraus operators / channels that have one to one functionality with Cirq ?

In python I can just do something like:

```python3
noisy_op = cirq.depolarize(0.001)(qubit)
```
and then place it inside of my circuit.

In qsim I can construct channels and kraus ops but it looks like I have to do things kind of by hand with "custom" gates. It would be nice to instead just do:

```C++
auto noisy_op = Depolarizing<>::Create(...)
```
Much like we do with Cirq's gates in `cirq_gates.h`.

95-martin-orion:
+1 - this seems like a nice feature to support our C++ qsim users.

In the meantime, Mike, you can take a look at our Cirq-to-qsim translations in PR #287. In particular:
- `translate_cirq_to_qtrajectory` in `qsim_circuit.py`
    - Extracts the key features of channels and mixtures, then sends them to `add_channel`
- `add_channel` in `pybind_main.cc`
    - Converts the features from `translate_cirq_to_qtrajectory` into a qsim Channel object

If you need a temporary workaround, you could duplicate portions of those methods for the necessary behavior.

MichaelBroughton:
Thx. Left a few general comments on things. I think in our case it'd be worth waiting for a a nice way to create the 1:1 Cirq channels and get this solved. Doing the workarounds on gates back in the day and then moving onto `cirq_gates.h` once they were added wasn't a ton of fun so I think we might wait until this is in, we're in no rush on this feature for TFQ until there's a clean way to do it :)

PR #300, opened 2021-03-02
Update version to 0.8.0

95-martin-orion:
Do not merge until after #299.

The primary motivation for this release is to avoid incompatibility with Cirq due to #278; however, enough features have been submitted since 0.7.1 that a new minor version is justified.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ description = ('Schrödinger and Schrödinger-Feynman simulators for quantum cir
 # README file as long_description.
 long_description = open('README.md', encoding='utf-8').read()
 
-__version__ = '0.7.1'
+__version__ = '0.8.0'
 
 setup(
     name='qsimcirq',


MichaelBroughton:
Would this be able to hold off until a fix for #288 is merged in as well ? We need that feature to progress with noise in TFQ and @sergeisakov mentioned it would be coming before this week is out.

95-martin-orion:
> Would this be able to hold off until a fix for #288 is merged in as well?

The upcoming Cirq release (scheduled for end of this week) is incompatible with the latest qsimcirq release due to a disagreement in `_resolve_parameters_`. Without this release, `import cirq, qsimcirq` will misbehave for parameterized circuits.

As such, this release is relatively urgent. I'm fine with cutting another release for the #288 fix, even if it's shortly afterwards - qsim releases are pretty cheap.

MichaelBroughton:
Sounds good. Interestingly enough we got burned by that exact same change.

MichaelBroughton (approved):

PR #301, opened 2021-03-03
Reduce proliferation of methods in pybind

95-martin-orion:
There are eight different simulation "types" supported in the pybind layer, resulting from three binary choices in simulation:
- input state or vector,
- noisy or noiseless simulation,
- return states or expectation values

The amount of code repetition in this file was starting to get silly, so I've created SimulatorHelper to help. Crucially, this class separates state prep, simulation, and output into different steps so each simulation type only does the work that's absolutely required for it.

Prerequisite for expectation value improvements.

PR #303, opened 2021-03-04
Support aggregation of expectation values in qsimcirq

### LINKED ISSUE #293: Align noisy simulations in qsimcirq with API expectations
Identified in #292. Repeated simulation of a circuit is often necessary when that circuit has non-unitary intermediate gates, whether those gates represent measurements or noise. Currently, qsimcirq has no special handling for this behavior; each simulation is entirely ignorant of previous simulations.

By either preserving information in qsimcirq objects (QSimSimulator and QSimCircuit) or providing alternate interfaces for repeated execution, we can reduce the overhead from things like Cirq-to-qsim circuit translation.

95-martin-orion:
Fixes #293.

Users can now specify a "noisy repetitions" option for `QSimSimulator`. This value determines how many times to execute a noisy circuit when calculating expectation values for that circuit.

### DIFF
--- a/pybind_interface/pybind_main.cpp
+++ b/pybind_interface/pybind_main.cpp
@@ -482,10 +482,37 @@ class SimulatorHelper {
                             unsigned>>& opsums_and_qubit_counts,
       bool is_noisy, const StateType& input_state) {
     auto helper = SimulatorHelper(options, is_noisy);
-    if (!helper.is_valid || !helper.simulate(input_state)) {
+    if (!helper.is_valid) {
       return {};
     }
-    return helper.get_expectation_value(opsums_and_qubit_counts);
+    if (!is_noisy) {
+      if (!helper.simulate(input_state)) {
+        return {};
+      }
+      return helper.get_expectation_value(opsums_and_qubit_counts);
+    }
+
+    // Aggregate expectation values for noisy circuits.
+    std::vector<std::complex<double>> results(
+      opsums_and_qubit_counts.size(), 0);
+    For aggregator(helper.num_threads);
+    auto add_evs = [&results](unsigned n, unsigned m, uint64_t i,
+                              const std::vector<std::complex<double>>& evs) {
+      results[i] += evs.at(i);
+    };
+    for (unsigned rep = 0; rep < helper.noisy_reps; ++rep) {
+      if (!helper.simulate(input_state)) {
+        return {};
+      }
+      auto evs = helper.get_expectation_value(opsums_and_qubit_counts);
+      aggregator.Run(evs.size(), add_evs, evs);
+    }
+    auto avg_evs = [&results](unsigned n, unsigned m, uint64_t i,
+                              unsigned num_reps) {
+      results[i] /= num_reps;
+    };
+    aggregator.Run(results.size(), avg_evs, helper.noisy_reps);
+    return results;
   }
 
  private:
@@ -498,6 +525,7 @@ class SimulatorHelper {
       if (is_noisy) {
         ncircuit = getNoisyCircuit(options);
         num_qubits = parseOptions<unsigned>(options, "n\0");
+        noisy_reps = parseOptions<unsigned>(options, "r\0");
       } else {
         circuit = getCircuit(options);
         num_qubits = circuit.num_qubits;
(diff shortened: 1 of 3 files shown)


95-martin-orion:
TODO: rebase to #301 after that PR is merged.

95-martin-orion:
> TODO: rebase to #301 after that PR is merged.

Done. Ready for review.

pybind_interface/pybind_main.cpp
```diff
@@ -482,10 +482,37 @@ class SimulatorHelper {
...
+    auto add_evs = [&results](unsigned n, unsigned m, uint64_t i,
+                              const std::vector<std::complex<double>>& evs) {
+      results[i] += evs.at(i);
+    };
+    for (unsigned rep = 0; rep < helper.noisy_reps; ++rep) {
+      if (!helper.simulate(input_state)) {
+        return {};
+      }
+      auto evs = helper.get_expectation_value(opsums_and_qubit_counts);
+      aggregator.Run(evs.size(), add_evs, evs);
```
sergeisakov, line 508: It's probably better to use just the standard sequential `for` here and below. The parallel `for` will have advantage only for `opsums_and_qubit_counts` > 10000 or so.

pybind_interface/pybind_main.cpp
```diff
@@ -482,10 +482,37 @@ class SimulatorHelper {
...
+      return helper.get_expectation_value(opsums_and_qubit_counts);
+    }
+
+    // Aggregate expectation values for noisy circuits.
+    std::vector<std::complex<double>> results(
+      opsums_and_qubit_counts.size(), 0);
+    For aggregator(helper.num_threads);
+    auto add_evs = [&results](unsigned n, unsigned m, uint64_t i,
+                              const std::vector<std::complex<double>>& evs) {
+      results[i] += evs.at(i);
```
sergeisakov, line 501: What is the reason for using `evs.at(i)` instead of `evs[i]`?

pybind_interface/pybind_main.cpp
```diff
@@ -482,10 +482,37 @@ class SimulatorHelper {
...
+    };
+    for (unsigned rep = 0; rep < helper.noisy_reps; ++rep) {
+      if (!helper.simulate(input_state)) {
+        return {};
+      }
+      auto evs = helper.get_expectation_value(opsums_and_qubit_counts);
+      aggregator.Run(evs.size(), add_evs, evs);
+    }
+    auto avg_evs = [&results](unsigned n, unsigned m, uint64_t i,
```
sergeisakov, line 510: I would replace that with
```C++
auto avg_evs = [&results](unsigned n, unsigned m, uint64_t i,
                          double inverse_num_reps) {
  results[i] *= inverse_num_reps;
};
aggregator.Run(results.size(), avg_evs, 1.0 / helper.noisy_reps);
```

qsimcirq/qsim_simulator.py
```diff
@@ -80,9 +80,13 @@ def __init__(self, qsim_options: dict = {},
             applied to all circuits run using this simulator. Accepted keys and
             their behavior are as follows:
                 - 'f': int (> 0). Maximum size of fused gates. Default: 2.
+                - 'r': int (> 0). Noisy repetitions (see below). Default: 1.
```
sergeisakov, line 83: To me passing the number of repetitions as a simulator option (these are simulator options not circuit options as stated a few lines above) doesn't make any sense. The number of repetitions is not a simulator option. It's an important parameter for the quantum trajectory method and that should be reflected in a higher level API. I think it should be passed as `repetitions` to `simulate_expectation_values_sweep` (or to a function with some better name). `simulate_expectation_values_sweep` with just one repetition doesn't really make a lot of sense for noisy simulations if the results get averaged in this function.

sergeisakov (changes requested):

pybind_interface/pybind_main.cpp

pybind_interface/pybind_main.cpp
95-martin-orion: I remember the compiler complaining about `evs` being `const`, but it seems fine now. Switched to `evs[i]`.

pybind_interface/pybind_main.cpp

sergeisakov (approved):

qsimcirq/qsim_simulator.py
95-martin-orion: This was suggested in my discussion with the Cirq team, with the primary reason being that setting "noisy repetitions" is analogous to setting simulation fidelity. Unless you're testing how different numbers of repetitions affect the quality of your result, you'll likely want a single value for this across multiple circuits.

The other motivating factor here is that `repetitions` doesn't make sense for most other `simulate_expectation_values` use cases (e.g. noiseless circuits or density matrix simulation). In the future, it might be worth the extra effort to have a separate method in Cirq (and we may revisit this then), but for the time being this is qsim-specific behavior.

PR #305, opened 2021-03-04
Add Cirq channels.

### LINKED ISSUE #288: Ensure 1:1 parity with Cirq channels

sergeisakov:
Closes #288

### DIFF
--- a/lib/BUILD
+++ b/lib/BUILD
@@ -8,6 +8,9 @@ cc_library(
     hdrs = [
         "bits.h",
         "bitstring.h",
+        "channel.h",
+        "channels_cirq.h",
+        "circuit_noisy.h",
         "circuit_qsim_parser.h",
         "circuit.h",
         "expect.h",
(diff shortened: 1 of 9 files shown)


MichaelBroughton (approved):
Looks great! Thx so much!

95-martin-orion:
Happy to defer to Michael on the review here, since I didn't see any issues in a quick pass. One question, though: is there any incentive to update qsimcirq to also use these new channel types?

sergeisakov:
> Happy to defer to Michael on the review here, since I didn't see any issues in a quick pass. One question, though: is there any incentive to update qsimcirq to also use these new channel types?

Yes, we can update qsimcirq to use these new channel types. We have a number of predefined gates and custom matrix gates in qsimcirq. Similarly, we can have a number of predefined channels and custom channels.

95-martin-orion:
> Yes, we can update qsimcirq to use these new channel types.

Is there any benefit to doing this, outside of readability? e.g. is performance better when using these channel types vs. just using the raw matrix-based Channels like we currently do in qsimcirq?

sergeisakov:
> Is there any benefit to doing this, outside of readability? e.g. is performance better when using these channel types vs. just using the raw matrix-based Channels like we currently do in qsimcirq?

There is no need to calculate SVD explicitly for each channel. This is probably the only benefit, outside of readability.

PR #310, opened 2021-03-10
Add quantum trajectory test.

sergeisakov:
Add a quantum trajectory test that shows how the final state can be uncomputed back to the initial state.


MichaelBroughton (approved):
Thanks! Appreciate the example of the use case via test.

PR #312, opened 2021-03-17
Refactor quantum trajectory simulator.

### LINKED ISSUE #311: Support other for objects in qtrajectory
In TensorFlow Quantum we have our own version of the parallelfor that uses TensorFlow's threadpool. Here's a snippet:
```C++
// Instantiate qsim objects.
const auto tfq_for = tfq::QsimFor(context);
using Simulator = qsim::Simulator<const tfq::QsimFor&>;
using StateSpace = Simulator::StateSpace;

// Begin simulation.
Simulator sim = Simulator(tfq_for);
.
.
.
```
Inside of [qtrajectory](https://github.com/quantumlib/qsim/blob/master/lib/qtrajectory.h#L199) there is:
```C++
 private:
  static bool RunIteration(uint64_t rep,
                           const Parameter& param, unsigned num_qubits,
                           typename NoisyCircuit<Gate>::const_iterator cfirst,
                           typename NoisyCircuit<Gate>::const_iterator clast,
                           std::vector<const Gate*>& gates,
                           State& scratch, State& state,
                           std::vector<uint64_t>& stat) {
    if (param.collect_kop_stat || param.collect_mea_stat) {
      stat.reserve(std::size_t(clast - cfirst));
      stat.resize(0);
    }

    StateSpace state_space(param.num_threads);
    Simulator simulator(param.num_threads);
```
Which doesn't allow the user to choose anything other than the number of threads they would like to supply to whatever comes out of `formux.h`. Could this behavior be changed to support TFQ so that a user could pass in their own `For` implementation @sergeisakov , @95-martin-orion ? Is there a way around this that I might be missing ?

sergeisakov:
Closes #311

### DIFF
--- a/pybind_interface/pybind_main.h
+++ b/pybind_interface/pybind_main.h
@@ -29,8 +29,6 @@ namespace py = pybind11;
 #include "../lib/gates_cirq.h"
 #include "../lib/qtrajectory.h"
 
-PYBIND11_MAKE_OPAQUE(qsim::NoisyCircuit<qsim::Cirq::GateCirq<float>>);
-
 // Methods for mutating noiseless circuits.
 void add_gate(const qsim::Cirq::GateKind gate_kind, const unsigned time,
               const std::vector<unsigned>& qubits,
--- a/pybind_interface/pybind_main.h
+++ b/pybind_interface/pybind_main.h
@@ -216,7 +216,10 @@ PYBIND11_MODULE(qsim, m) {
     .def_readwrite("num_qubits", &Circuit::num_qubits)
     .def_readwrite("gates", &Circuit::gates);
 
-  py::bind_vector<NoisyCircuit>(m, "NoisyCircuit");
+  py::class_<NoisyCircuit>(m, "NoisyCircuit")
+    .def(py::init<>())
+    .def_readwrite("num_qubits", &NoisyCircuit::num_qubits)
+    .def_readwrite("channels", &NoisyCircuit::channels);
 
   py::class_<OpString>(m, "OpString")
     .def(py::init<>())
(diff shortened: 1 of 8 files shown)


MichaelBroughton (approved):
Thanks!

pybind_interface/pybind_main.h
```diff
@@ -216,7 +216,10 @@ PYBIND11_MODULE(qsim, m) {
     .def_readwrite("num_qubits", &Circuit::num_qubits)
     .def_readwrite("gates", &Circuit::gates);
 
-  py::bind_vector<NoisyCircuit>(m, "NoisyCircuit");
+  py::class_<NoisyCircuit>(m, "NoisyCircuit")
```
95-martin-orion, line 219: This should allow us to remove the `PYBIND_MAKE_OPAQUE` line at the top of the file:
https://github.com/quantumlib/qsim/blob/6ad4cd5a3f34cc31012987b6f77ce3a564206f48/pybind_interface/pybind_main.h#L32

qsimcirq/qsim_simulator.py
```diff
@@ -88,13 +88,13 @@ def __init__(self, qsim_options: dict = {},
             over when calculating expectation values for a noisy circuit.
             Note that this does not apply to other simulation types.
         seed: A random state or seed object, as defined in cirq.value.
-    
+
     Raises:
-        ValueError if internal keys 'c' or 'i' are included in 'qsim_options'.
+        ValueError if internal keys 'c', 'i' or 's'are included in 'qsim_options'.
```
95-martin-orion, line 93: "or 's'are" -> "or 's' are"

95-martin-orion (changes requested):
Couple of small nitpicks.

Thank you for the multitude of qsimcirq cleanup items in this PR as well! I think the separation of names in `qtrajectory` will help as well.

pybind_interface/pybind_main.h
sergeisakov: Done.

qsimcirq/qsim_simulator.py
sergeisakov: Done.

95-martin-orion (approved):

PR #314, opened 2021-03-23
Add deps required by internal version

95-martin-orion:
These dependencies are required for the internal version of these BUILD rules.

(long story short, internal bazel is not a fan of transitive dependencies)

### DIFF
--- a/lib/BUILD
+++ b/lib/BUILD
@@ -469,8 +469,7 @@ cc_library(
     hdrs = ["unitary_calculator_avx.h"],
     deps = [
         ":bits",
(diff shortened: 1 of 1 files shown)


sergeisakov (approved):
Thank you. That was a typo. Probably dependence on `unitaryspace_basic` in lines 473 and 491 is not required.

95-martin-orion:
> Thank you. That was a typo. Probably dependence on `unitaryspace_basic` in lines 473 and 491 is not required.

Confirmed on internal version - removed unnecessary deps.

PR #317, opened 2021-03-30
Update index code snippet.

95-martin-orion:
CC @karlunho, @dstrain115

Discussion with users highlighted two issues with the previous code snippet:
- The 6x6 grid is too large for the average desktop (requires ~256GB RAM)
- Lack of measurements in the circuit triggered a Cirq error

My concerns with this:
- Will changing the snippet length mess up the [qsim landing page](https://quantumai.google/qsim)?
- Was there a reason we needed to mention multi-core Intel CPUs?

### DIFF
--- a/docs/_index.yaml
+++ b/docs/_index.yaml
@@ -59,7 +59,7 @@ landing_page:
         import cirq
         import qsimcirq
 
-        # Pick up to ~25 qubits to simulate (requires ~128MB of RAM)
+        # Pick up to ~25 qubits to simulate (requires ~256MB of RAM)
         qubits = [cirq.GridQubit(i,j) for i in range(5) for j in range(5)]
 
         # Define a circuit to run
--- a/docs/_index.yaml
+++ b/docs/_index.yaml
@@ -59,13 +59,17 @@ landing_page:
         import cirq
         import qsimcirq
 
-        # Pick up to ~36 qubits to run on a large multi-core Intel CPU machine
-        qubits = [cirq.GridQubit(i,j) for i in range(6) for j in range(6)]
+        # Pick up to ~25 qubits to simulate (requires ~128MB of RAM)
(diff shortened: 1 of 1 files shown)


dstrain115:
It's fine with me.  I think the snippet should be a "hello world" that's easy to run and works (not one that requires a multi-processor setup).

95-martin-orion:
+@lamberta for feedback on how this will affect the site. Will the code box resize automatically, or do we need to manually adjust it to fit the new snippet?

lamberta:
The snippet box will automatically expand *vertically*. But the max width is capped so it will add a horizontal scrollbar if the content extends beyond the visible area.

We generally use these landing page snippets to illustrate very high-level usage or concepts—kind of a way to let the reader know what problems this library will help them solve and confirm they're in the right section of the website. Correct is good, but it's also fine to skip over setup that detracts from that high-level messaging. You can always link to a working Colab tutorial for that.

95-martin-orion:
Width shouldn't be an issue - the font is fixed-width and no lines were made longer by this change.

My reasoning for updating this is that a user was trying it for their first use of qsim. From their perspective that's a logical move, since it's the first code snippet they encounter on the site.

95-martin-orion:
+@sergeisakov for final review and merge.

docs/_index.yaml
```diff
@@ -59,14 +59,18 @@ landing_page:
         import cirq
         import qsimcirq
 
-        # Pick up to ~36 qubits to run on a large multi-core Intel CPU machine
-        qubits = [cirq.GridQubit(i,j) for i in range(6) for j in range(6)]
+        # Pick up to ~25 qubits to simulate (requires ~128MB of RAM)
```
sergeisakov, line 62: It requires 256MB of RAM for 25 qubits.

sergeisakov (approved):

dstrain115 (approved):

PR #322, opened 2021-04-20
Add support for avx512.

sergeisakov:


95-martin-orion (changes requested):
I was able to reproduce the passing tests with both `make run-cxx-tests` and `bazel test tests:all` on a machine with AVX512 support.

Is it possible to preserve the bazel BUILD rules for the new tests and skip them on machines without AVX512 support?

95-martin-orion:
The "passing" kokoro test for the previous run had a failure on Windows - opened #329 to track.

95-martin-orion:
Discussed and thoroughly tested - given that CI machines lack AVX512 support, I don't see a good way to keep the tests in this PR. I'm willing to approve this PR with the AVX512 test BUILD rules removed as long as we include notes in each of those tests that they are not covered by CI.

PR #324, opened 2021-04-21
Support iterable qubit_order.

### LINKED ISSUE #323: Setting qubit_order to reverse(qubit_list) doesn't work for QSimSimulator
Running the following

```python
q0, q1 = cirq.LineQubit.range(2)

circuit = cirq.Circuit(
    cirq.H(q0), cirq.H(q1),
)
qsim_simulator = qsimcirq.QSimSimulator()
results = qsim_simulator.simulate(circuit, qubit_order=reversed([q1, q0]))
```
yields `ValueError: Unexpected extra qubits: {cirq.LineQubit(0), cirq.LineQubit(1)}.` However `list(reversed([q1, q0]))` works. An Iterable type isn't handled correctly for some reason it seems like.

Not wrapping it in a `list()` works for `cirq.Simulator()` and `cirq.DensityMatrixSimulator()`.

The expected type for `qubit_order` is `cirq.ops.QubitOrderOrList`, which is
```
QubitOrderOrList = Union[
    cirq.ops.QubitOrder,
    collections.abc.Iterable[cirq.ops.Qid]
]
```
so any Iterable on qubits *should* work.

The only difference I can tell so far between cirq simulators and QSimSimulator is that cirq simulators call `as_qubit_order()` twice:
https://github.com/quantumlib/Cirq/blob/574652c9defe87541ee2a8725d8d09da544b5ed7/cirq-core/cirq/sim/simulator.py#L425
then
https://github.com/quantumlib/Cirq/blob/574652c9defe87541ee2a8725d8d09da544b5ed7/cirq-core/cirq/sim/simulator.py#L536

but that shouldn't have done anything...
https://github.com/quantumlib/Cirq/blob/574652c9defe87541ee2a8725d8d09da544b5ed7/cirq-core/cirq/ops/qubit_order.py#L123

95-martin-orion:
Fixes #323. The input `qubit_order` argument can only be consumed once, as the simulator API allows Iterable `qubit_order`s.

### DIFF
--- a/qsimcirq_tests/qsimcirq_test.py
+++ b/qsimcirq_tests/qsimcirq_test.py
@@ -215,6 +215,39 @@ def test_invalid_params():
     _ = qsim_simulator.simulate_sweep(circuit, params=prs)
 
 
+def test_iterable_qubit_order():
+  # Check to confirm that iterable qubit_order works in all cases.
+  q0, q1 = cirq.LineQubit.range(2)
+  circuit = cirq.Circuit(
+    cirq.H(q0), cirq.H(q1),
+  )
+  qsim_simulator = qsimcirq.QSimSimulator()
+
+  assert qsim_simulator.compute_amplitudes(
+    circuit,
+    bitstrings=[0b00, 0b01],
+    qubit_order=reversed([q1, q0]),
+  ) == qsim_simulator.compute_amplitudes(circuit, bitstrings=[0b00, 0b01])
+
+  assert (
+    qsim_simulator.simulate(circuit, qubit_order=reversed([q1, q0])) ==
+    qsim_simulator.simulate(circuit, qubit_order=reversed([q1, q0]))
+  )
+
+  assert qsim_simulator.simulate_expectation_values_sweep(
+    circuit,
+    observables=[cirq.X(q0) * cirq.Z(q1)],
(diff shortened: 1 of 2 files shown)


qsimcirq_tests/qsimcirq_test.py
```diff
@@ -215,6 +215,39 @@ def test_invalid_params():
...
+
+  assert qsim_simulator.compute_amplitudes(
+    circuit,
+    bitstrings=[0b00, 0b01],
+    qubit_order=reversed([q1, q0]),
+  ) == qsim_simulator.compute_amplitudes(circuit, bitstrings=[0b00, 0b01])
+
+  assert (
+    qsim_simulator.simulate(circuit, qubit_order=reversed([q1, q0])) ==
+    qsim_simulator.simulate(circuit, qubit_order=reversed([q1, q0]))
```
sergeisakov, line 234: Should that be just `qsim_simulator.simulate(circuit)` in line 234?

sergeisakov (approved):

qsimcirq_tests/qsimcirq_test.py
95-martin-orion: Yup, thanks for the catch!

PR #325, opened 2021-04-22
Add py.typed file and include in published package

maffoo:
This allows code that incorporates qsimcirq to understand the inline type annotations. This will require publishing a new version of qsimcirq.

### DIFF
--- a/qsimcirq/py.typed
+++ b/qsimcirq/py.typed
@@ -1,0 +1,1 @@
+# Marker file for PEP 561. This package uses inline types.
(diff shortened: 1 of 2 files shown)


95-martin-orion (approved):
This seems fine to me. It doesn't look like we've had any major changes since 0.9.0 - I can release this as 0.9.1 if you'd like.

95-martin-orion:
Correction: this would be 0.9.2 (0.9.1 was a C++-only release)

maffoo:
@95-martin-orion, yes, if you can publish a 0.9.2 version, that would be great!

PR #326, opened 2021-04-22
Update version to 0.9.2

95-martin-orion:
Version 0.9.1 was a C++-only release; the new version is 0.9.2.

This version will have both a Github and pypi release.


maffoo (approved):
LGTM

PR #335, opened 2021-04-27
Ensure all qubits are used in qsimcirq result

### LINKED ISSUE #327: QSimSimulator does not respect "idle" qubits
In the cirq simulator, the `qubit_order` parameter specifies the "basis" in which the state vector will be expressed, and this can contain extra "idle" qubits that do not run any gates. QSimSimulator returns a resulting state vector with the wrong shape and also raises an exception when trying to produce the string representation of the result:
```python
In [1]: import cirq; import qsimcirq

In [2]: qs = cirq.LineQubit.range(2)

In [3]: cirq.Simulator().simulate(cirq.Circuit(cirq.X(qs[0])**0.5), initial_state=0, qubit_order=qs)
Out[3]: 
measurements: (no measurements)
output vector: (0.5+0.5j)|00⟩ + (0.5-0.5j)|01⟩

In [4]: r = qsimcirq.QSimSimulator().simulate(cirq.Circuit(cirq.X(qs[0])**0.5), initial_state=0, qubit_order=qs)

In [5]: r.state_vector()
Out[5]: array([0.49999997+0.49999997j, 0.49999997-0.49999997j], dtype=complex64)

In [6]: r
Out[6]: ---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
~/.virtualenvs/pyle/lib/python3.8/site-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()

~/.virtualenvs/pyle/lib/python3.8/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    389                             meth = cls._repr_pretty_
    390                             if callable(meth):
--> 391                                 return meth(obj, self, cycle)
    392                         if cls is not object \
    393                                 and callable(cls.__dict__.get('__repr__')):

~/.virtualenvs/pyle/lib/python3.8/site-packages/cirq/sim/state_vector_simulator.py in _repr_pretty_(self, p, cycle)
    186             p.text('StateVectorTrialResult(...)')
    187         else:
--> 188             p.text(str(self))
    189 
    190     def __repr__(self) -> str:

~/.virtualenvs/pyle/lib/python3.8/site-packages/cirq/sim/state_vector_simulator.py in __str__(self)
    175         final = self.state_vector()
    176         if len([1 for e in final if abs(e) > 0.001]) < 16:
--> 177             state_vector = self.dirac_notation(3)
    178         else:
    179             state_vector = str(final)

~/.virtualenvs/pyle/lib/python3.8/site-packages/cirq/sim/state_vector.py in dirac_notation(self, decimals)
     94             A pretty string consisting of a sum of computational basis kets
     95             and non-zero floats of the specified accuracy."""
---> 96         return qis.dirac_notation(self.state_vector(), decimals, qid_shape=self._qid_shape)
     97 
     98     def density_matrix_of(self, qubits: List[ops.Qid] = None) -> np.ndarray:

~/.virtualenvs/pyle/lib/python3.8/site-packages/cirq/qis/states.py in dirac_notation(state_vector, decimals, qid_shape)
    468     for x in range(len(perm_list)):
    469         format_str = "({:." + str(decimals) + "g})"
--> 470         val = round(state_vector[x].real, decimals) + 1j * round(state_vector[x].imag, decimals)
    471 
    472         if round(val.real, decimals) == 0 and round(val.imag, decimals) != 0:

IndexError: index 2 is out of bounds for axis 0 with size 2
```

### LINKED ISSUE #330: cirq.IdentityGate only supported up to 6 qubits
Seems like we should be able to support identity gates of any size by treating them as "no-ops".

95-martin-orion:
This PR addresses two closely-related issues.

Fixes #327:

The root issue here was that we sometimes used `Circuit.all_qubits()` for `num_qubits`, when `ordered_qubits` could sometimes have more qubits (e.g. if the user wanted "extra" qubits to appear in the result).

Fixes #330:

Identities will now decompose to no-ops. Since the qubits used by qsim are based on the input circuit + qubit order, this can be done safely without affecting the simulator behavior.

### DIFF
--- a/qsimcirq_tests/qsimcirq_test.py
+++ b/qsimcirq_tests/qsimcirq_test.py
@@ -67,7 +67,7 @@ def test_cirq_too_big_gate():
     qsimSim.compute_amplitudes(cirq_circuit, bitstrings=[0b0, 0b1])
 
 
-def test_cirq_too_big_identity():
+def test_cirq_giant_identity():
   # Pick qubits.
   a, b, c, d, e, f, g, h = [
       cirq.GridQubit(0, 0),
@@ -86,9 +86,13 @@ def test_cirq_too_big_identity():
     cirq.X(h),
   )
 
+  no_id_circuit = cirq.Circuit(cirq.X(h))
   qsimSim = qsimcirq.QSimSimulator()
-  with pytest.warns(RuntimeWarning, match='Identities on 7\+ qubits'):
-    qsimSim.compute_amplitudes(cirq_circuit, bitstrings=[0b0, 0b1])
+
+  assert (
+    qsimSim.simulate(cirq_circuit) ==
+    qsimSim.simulate(no_id_circuit, qubit_order=[a,b,c,d,e,f,g,h])
+  )
 
 
 @pytest.mark.parametrize('mode', ['noiseless', 'noisy'])
--- a/qsimcirq_tests/qsimcirq_test.py
+++ b/qsimcirq_tests/qsimcirq_test.py
@@ -32,6 +32,11 @@ class NoiseTrigger(cirq.SingleQubitGate):
     return (np.asarray([1, 0, 0, 1]),)
 
 
+def test_empty_circuit():
+  result = qsimcirq.QSimSimulator().simulate(cirq.Circuit())
+  assert result.final_state_vector.shape == (1,)
+
+
 def test_cirq_too_big_gate():
   # Pick qubits.
   a, b, c, d, e, f, g = [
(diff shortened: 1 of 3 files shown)


qsimcirq_tests/qsimcirq_test.py
```diff
@@ -248,6 +291,22 @@ def test_iterable_qubit_order():
   )
 
 
+@pytest.mark.parametrize('mode', ['noiseless', 'noisy'])
+def test_preserve_qubits(mode: str):
```
maffoo, line 295: nit: use a bool
```suggestion
@pytest.mark.parametrize('add_noise', [False, True])
def test_preserve_qubits(add_noise: bool):
```
then `if add_noise:`

maffoo (approved):
Minor comment, then LGTM

qsimcirq_tests/qsimcirq_test.py
95-martin-orion: Using a string here is intentional. It ensures that the test shows up as `test_preserve_qubits[noisy]` in pytest logs, instead of the more cryptic `test_preserve_qubits[True]`.

sergeisakov (approved):
LGTM

PR #341, opened 2021-05-06
Fix controlled-gate qubit limit.

95-martin-orion:
The error for too many qubits in a controlled gate suggests that the intent was to limit the number of _target_ qubits to 4, not the number of _total_ qubits. This PR brings the code in line with this description.


sergeisakov (approved):
LGTM

PR #343, opened 2021-05-06
Add CI check for Python formatting.

### LINKED ISSUE #328: Format python code with black?
The qsimcirq code looks like it follows internal google formatting style (two-space indent is a dead giveaway :-) ). I think it would be nice to use [black](https://github.com/psf/black) for formatting to be consistent with other open source projects, such as cirq itself (see https://github.com/quantumlib/Cirq/blob/master/check/format-incremental#L127). Setting up some infrastructure for autoformat the code and check it with a github action would also be nice. Does this make sense?

95-martin-orion:
This check should only trigger on changes to `*.py` files. Fixes #328 (for real this time).

### DIFF
--- a/.github/workflows/python_format.yml
+++ b/.github/workflows/python_format.yml
@@ -1,0 +1,24 @@
+on:
+  push:
+    paths:
+      # Only check formatting when python files are changed.
+      - '**/*.py'
+
+jobs:
+  format:
+    name: Format check
+    runs-on: ubuntu-16.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v1
+        with:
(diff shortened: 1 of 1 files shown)


maffoo (approved):
LGTM

.github/workflows/python_format.yml
```diff
@@ -0,0 +1,24 @@
+on:
+  push:
```
maffoo, line 2: Will this run on every push even if not part of a PR?

95-martin-orion: Good point. Updated to trigger only push to master, or on PRs to any branch (still only when python files change).

95-martin-orion: Demonstrated desired behavior in #344.

PR #344, opened 2021-05-06
Test commit for formatting CI checks

95-martin-orion:
Testing the new CI checks (and also whether CI changes can be tested this way!)


95-martin-orion:
Test failure confirms that the new CI check works!

PR #348, opened 2021-05-19
Update Ubuntu version in Github Actions

95-martin-orion:
The `ubuntu-16.04` environment is being turned down: https://github.com/actions/virtual-environments/issues/3287

Once we confirm that this conversion is OK for TFQ, we will switch qsim over as well. Note that the python formatting workflow is not tied to TFQ and can therefore use the latest Ubuntu version without loss of generality.


95-martin-orion:
From @MichaelBroughton: the TFQ version is currently locked to the TFQ toolchain [here](https://github.com/tensorflow/tensorflow/tree/master/third_party/toolchains/preconfig).

MichaelBroughton:
xref: https://github.com/tensorflow/tensorflow/tree/master/third_party/toolchains/preconfig
TFQ relies on this toolchain for most builds and it is 16.04.

MichaelBroughton (approved):
As of TF 2.5.0, the toolchains have been upgraded and everyone is using 1 now. This should be good to go!

Issue #346, opened 2021-05-13
Windows Python wheel fails with heap corruption exception


laurynasas:
As part of #313 issue we are building a PyPI wheel for Windows platform based on Github Actions Windows Server 2019 virtual environment. The wheel is successfully built using `python3 setup.py sdist bdist_wheel` that utilizes `cmake-3.20` and for building and `MSVC 14.28.29910` for compilation of the underlying libraries. The build also uses `pybind11` to create binding between Python and C++ libraries.

Installing the built PyPI wheel on Windows Server 2019 virtual environment and running pytests:
```
pytest qsimcirq_tests/qsimcirq_test.py -k "test_cirq_qsim_run"
``` 

fails with the following stack trace:
```
============================= test session starts =============================
platform win32 -- Python 3.7.9, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: D:\a\qsim\qsim
collected 51 items / 49 deselected / 2 selected

Windows fatal exception: code 0xc0000374

Current thread 0x0000199c (most recent call first):
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\qsimcirq\qsim_simulator.py", line 128 in _run
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\cirq\sim\simulator.py", line 103 in run_sweep
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\cirq\work\sampler.py", line 49 in run
  File "D:\a\qsim\qsim\qsimcirq_tests\qsimcirq_test.py", line 245 in test_cirq_qsim_run
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\_pytest\python.py", line 183 in pytest_pyfunc_call
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\pluggy\callers.py", line 187 in _multicall
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\pluggy\manager.py", line 87 in <lambda>
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\pluggy\manager.py", line 93 in _hookexec
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\pluggy\hooks.py", line 286 in __call__
[...]

PR #355, opened 2021-05-25
Fix how memory of the Numpy array created by the C libs is deallocated

laurynasas:
Add pybind11 capsule fix for deallocating Numpy array as per [discussion](https://github.com/quantumlib/qsim/issues/346).


sergeisakov (approved):
LGTM

PR #356, opened 2021-05-26
Disable avx512 asan tests.

sergeisakov:


95-martin-orion (approved):
Should we revert this when #348 is eventually merged, or is this a better long-term solution for the test failures?

sergeisakov:
I think we should revert this when #348 is eventually merged.

Issue #329, opened 2021-04-26
Kokoro builds sometimes ignore failures


95-martin-orion:
Seen in #322, where the "passing" Kokoro build links to the MacOS run (which passed), but the Windows build failed.

My best guess for now is that Kokoro determines pass/fail based on the first completed test. It seems like there are some tools for chaining/grouping Kokoro tests; I'll look into those.

95-martin-orion:
A fix was rolled out in the internal Kokoro pipeline and verified in #359 (no Github changes required).

PR #358, opened 2021-05-27
Fix kokoro failure.

sergeisakov:


95-martin-orion (approved):
Thanks for the quick fix! I'll post an update on this PR when the kokoro build completes.

This also prompted me to resolve #329 - I'm working on a Google-internal change for that now.

95-martin-orion:
I missed that you had already run Kokoro - this is good to go.

PR #359, opened 2021-05-27
[TEST] Reinstate Kokoro failure

95-martin-orion:
This PR is purely for testing ongoing changes to the Kokoro workflow. **DO NOT SUBMIT**.


95-martin-orion:
These tests demonstrate that the new config works as expected:
- If all tests pass, the kokoro check passes.
- If **any** test fails, the kokoro check fails.

Issue #313, opened 2021-03-19
Use wheels in pypi packaging


95-martin-orion:
The goal of this issue is to resolve #242 and #259.

Currently, `pip install qsimcirq` requires building qsim locally. This is fine for Linux, but (as seen in the issues above) often encounters difficulty on MacOS and Windows. In order to resolve this, we need to replace the current packaging with a portable wheel as described by [PEP 513](https://www.python.org/dev/peps/pep-0513/).

95-martin-orion:
Comment from @Strilanc: cross-platform wheels may be tricky. Consider looking into `cibuildwheel`, which is how [Stim](https://github.com/quantumlib/Stim) builds for Mac and Windows.

Strilanc:
Stim doesn't use cibuildwheels to make the wheels, it uses cibuildwheels to verify that the source compiles on other platforms. You *can* use it to build the wheels, but I'm not confident they will generalize beyond the build machines when AVX instructions are involved.

95-martin-orion:
This has been resolved.

PR #360, opened 2021-06-08
Create base MacOS CI pipeline

laurynasas:
This PR addresses a cross-platform wheels [issue](https://github.com/quantumlib/qsim/issues/313) and adds support for building and uploading MacOS wheels for `cp36, cp37, cp38, cp39` on `x86_64` architecture.


95-martin-orion:
From the checks this appears to be good to go. You can auto-apply formatting fixes with `check/format-incremental --apply`.

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,58 @@
...
+        run: bash build_tools/test_libs.sh
+
+      - name: Build wheels
+        run: python -m cibuildwheel --output-dir wheelhouse
+
+      - uses: actions/upload-artifact@v2
+        with:
+          path: ./wheelhouse/*.whl
+
+      - name: Publish package
```
95-martin-orion, line 49: It seems like the preferred way of doing this is to have a separate workflow with the [`on: release` trigger](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration).

build_tools/test_libs.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -e  # fail and exit on any command erroring
+set -x  # print evaluated commands
+
+which bazel
+bazel version
+
+# Attempt to build all components in SSE and basic mode.
+# The Kokoro MacOS VMs are not configured for AVX2 or OpenMP, so these modes
+# are excluded from the build and test process.
```
95-martin-orion, line 10: Update comments to refer to Github Actions, rather than Kokoro.

95-martin-orion (changes requested):

.github/workflows/wheels.yml
laurynasas: I know it's possible to share data between jobs but how is it possible to persist built wheels between workflows?

95-martin-orion: Instead of persisting the wheel between workflows, we can just rebuild the wheel in the release workflow. (I usually cut releases via the Github API, which IIUC separates the "release" and "push" events.)

laurynasas: Yeah, that makes sense, I opened a new PR with the changes https://github.com/quantumlib/qsim/pull/379

PR #364, opened 2021-06-21
Prepare tests for CUDA simulator.

sergeisakov:


sergeisakov:
> This all seems reasonable to me. For my own understanding: what purpose does the `Factory` object serve in the CUDA simulator? I imagine it has to do with coordinating simulator creation across the GPU cores?

The `Factory` object helps to construct the simulator and state space objects. The constructors of the CPU simulator and the CPU state space take parameters that are used to construct the `For` object. The CUDA simulator and the CUDA state space don't depend on such a `For` object. There are other parameters. So, say, `Simulator simulator(1)` is not a good way to construct the CUDA simulator.

PR #365, opened 2021-06-21
MPS Implementation p1.

MichaelBroughton:
First step in adding the MPS design discussed in the simulator meetings. This PR adds the bare bones `mps_statespace.h` file. Once this is reviewed and in, the next step will be to put in `mps_simulator.h` (Which I'm planning to use Eigen3 for instead of writing out the intrinsics by hand) starting with single qubit functionality + tests and then building up to two qubit functionality + tests.

### DIFF
--- a/lib/mps_statespace.h
+++ b/lib/mps_statespace.h
@@ -83,16 +83,16 @@ class MPSStateSpace {
   static MPS CreateMPS(unsigned num_qubits, unsigned bond_dim) {
     auto end_sizes = 2 * 4 * bond_dim;
     auto internal_sizes = 4 * bond_dim * bond_dim * num_qubits;
-    // use two extra "internal style" blocks past the end of the
+    // Use two extra "internal style" blocks past the end of the
     //   working allocation for scratch space. Needed for gate
     //   application.
     auto size = sizeof(fp_type) * (end_sizes + internal_sizes);
 
 #ifdef _WIN32
     Pointer ptr{(fp_type*)_aligned_malloc(size, 64), &detail::free};
-    bool is_null = ptr.get() != nullptr return MPS{std::move(ptr),
-                                                   is_null ? num_qubits : 0,
-                                                   is_null ? bond_dim : 0};
+    bool is_null = ptr.get() != nullptr;
+    return MPS{std::move(ptr), is_null ? num_qubits : 0,
+               is_null ? bond_dim : 0};
 #else
     void* p = nullptr;
     if (posix_memalign(&p, 64, size) == 0) {
@@ -122,7 +122,7 @@ class MPSStateSpace {
   }
 
   // Copies the state contents of one MPS to another.
-  //  ignores scratch data.
+  //  Ignores scratch data.
   bool CopyMPS(const MPS& src, MPS& dest) const {
     if ((src.num_qubits() != dest.num_qubits()) ||
         src.bond_dim() != dest.bond_dim()) {
--- a/lib/mps_statespace.h
+++ b/lib/mps_statespace.h
@@ -122,7 +122,7 @@ class MPSStateSpace {
   }
 
   // Copies the state contents of one MPS to another.
-  //  Ignores scratch data.
+  // Ignores scratch data.
   bool CopyMPS(const MPS& src, MPS& dest) const {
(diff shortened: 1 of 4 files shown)


MichaelBroughton:
@95-martin-orion Was that CI failure me or was it already breaking ?

lib/mps_statespace.h
```diff
@@ -0,0 +1,154 @@
...
+    auto end_sizes = 2 * 4 * bond_dim;
+    auto internal_sizes = 4 * bond_dim * bond_dim * num_qubits;
+    // use two extra "internal style" blocks past the end of the
+    //   working allocation for scratch space. Needed for gate
+    //   application.
+    auto size = sizeof(fp_type) * (end_sizes + internal_sizes);
+
+#ifdef _WIN32
+    Pointer ptr{(fp_type*)_aligned_malloc(size, 64), &detail::free};
+    bool is_null = ptr.get() != nullptr return MPS{std::move(ptr),
```
sergeisakov, line 93: Should the return statement be on a separate line?

lib/mps_statespace.h
```diff
@@ -0,0 +1,154 @@
...
+  // Get the pointer offset to the beginning of an MPS block.
+  unsigned GetBlockOffset(const MPS& state, unsigned i) const {
+    if (i == 0) {
+      return 0;
+    }
+    return 4 * state.bond_dim() * (1 + state.bond_dim() * (i - 1));
+  }
+
+  // Copies the state contents of one MPS to another.
+  //  ignores scratch data.
```
sergeisakov, line 125: Remove space and make "i" capital?

lib/mps_statespace.h
```diff
@@ -0,0 +1,154 @@
...
+
+  // Note: ForArgs are currently unused.
+  template <typename... ForArgs>
+  MPSStateSpace(ForArgs&&... args) : for_(args...) {}
+
+  // Requires num_qubits >= 2 and bond_dim >= 2.
+  static MPS CreateMPS(unsigned num_qubits, unsigned bond_dim) {
+    auto end_sizes = 2 * 4 * bond_dim;
+    auto internal_sizes = 4 * bond_dim * bond_dim * num_qubits;
+    // use two extra "internal style" blocks past the end of the
```
sergeisakov, line 86: Make 'u' capital?

lib/mps_statespace.h
```diff
@@ -0,0 +1,154 @@
...
+#endif
+  }
+
+  unsigned Size(const MPS& state) const {
+    auto end_sizes = 2 * 4 * state.bond_dim();
+    auto internal_sizes = 4 * state.bond_dim() * state.bond_dim();
+    return end_sizes + internal_sizes * (state.num_qubits() - 2);
+  }
+
+  unsigned RawSize(const MPS& state) const {
```
sergeisakov, line 112: `unsigned` is sufficient to store all the sizes and offsets, right?

sergeisakov (changes requested):
Looks good to me. I just have a few minor suggestions.

lib/mps_statespace.h
MichaelBroughton: I think so. Rolling with unsigned types lets you address 2**32 - 1 places. Which is like 15+ Gigabytes worth of floats. I think the computational complexity of MPS being `n^3` for multiplications and SVDs vs `n^2` on memory you'd end up waiting an eternity to diagonalize such big matrices where memory wouldn't really be the main issue. Would you feel better if I bumped up to `long` ?

lib/mps_statespace.h
MichaelBroughton: Oops. Somehow that still worked with the missing semicolon. Fixed.

lib/mps_statespace.h
MichaelBroughton: Done.

lib/mps_statespace.h
MichaelBroughton: Done.

lib/mps_statespace.h
sergeisakov: If `unsigned` is sufficient then you don't need to use `long`.

sergeisakov (approved):
LGTM

Issue #354, opened 2021-05-25
Clean up qsim import hierarchy


95-martin-orion:
It's been pointed out that the use of `from ${MODULE} import *` in the `qsimcirq/__init__.py` file is not recommended. We should clean these up to behave more like the Cirq import hierarchy. Requirements for this:

- Only "public" classes and methods should be exposed through `__init__.py`
- Users should be able to address public qsimcirq objects as `qsimcirq.<object>`, instead of `qsimcirq.<module>.<object>`
- We should avoid breaking existing behavior as much as possible

PR #367, opened 2021-06-23
Replace star imports to fix mypy issue.

### LINKED ISSUE #349: Typechecking qsimcirq fails intermittently
We've been having intermittent failures when running mypy to typecheck our code which uses qsimcirq:
```
/usr/local/google/home/<user>/.virtualenvs/pyle/lib/python3.8/site-packages/qsimcirq/__init__.py:1: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
If this issue continues with mypy master, please report a bug at https://github.com/python/mypy/issues
version: 0.812
/usr/local/google/home/quintanac/.virtualenvs/pyle/lib/python3.8/site-packages/qsimcirq/__init__.py:1: : note: please use --show-traceback to print a traceback when reporting a bug
```
This puts mypy in a bad state that can only be recovered by deleting mypy's cache directory. The full traceback doesn't reveal much:
```
/usr/local/google/home/<user>/.virtualenvs/pyle/lib/python3.8/site-packages/qsimcirq/__init__.py:1: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 0.812
Traceback (most recent call last):
  File "mypy/semanal.py", line 4835, in accept
  File "mypy/nodes.py", line 396, in accept
  File "mypy/semanal.py", line 1949, in visit_import_all
  File "mypy/semanal.py", line 4491, in add_imported_symbol
  File "mypy/semanal.py", line 4396, in add_symbol_table_node
  File "mypy/semanal.py", line 4554, in defer
AssertionError: Must not defer during final iteration
/usr/local/google/home/quintanac/.virtualenvs/pyle/lib/python3.8/site-packages/qsimcirq/__init__.py:1: : note: use --pdb to drop into pdb
```
There may be a legitimate mypy bug at play, but it might be worth looking at the type annotations in qsimcirq to see if there's anything suspicious. I wonder if if could be getting confused by the wildcard imports in `__init__.py`, for example, which are considered bad practice.

kunalq:
Currently, import `*` from these modules breaks `mypy`, either because `__all__` is missing or due to another internal bug.

Aside from fixing the mypy issue, replacing these with explicit exports is better practice anyway as it delineates a public vs private API.

Note that this is a subtle functionality change if anyone was relying on internal methods such as `qsim_simulator._needs_trajectories` or if they were relying on cirq/numpy imports via the `qsim` namespace. I feel strongly that we should discourage both use cases, but I can be convinced to add a temporary shim.

Fixes #349

### DIFF
--- a/qsimcirq/__init__.py
+++ b/qsimcirq/__init__.py
@@ -1,5 +1,3 @@
-from .qsim_circuit import *
-from .qsim_simulator import *
-from .qsimh_simulator import *
-
-
+from .qsim_circuit import add_op_to_opstring, add_op_to_circuit, QSimCircuit
+from .qsim_simulator import QSimSimulatorState, QSimSimulatorTrialResult, QSimSimulator
+from .qsimh_simulator import QSimhSimulator


95-martin-orion:
This will also fix #354. I don't see an issue with breaking workflows that depended on use of "private" methods in qsim, so no temporary shim is necessary.

Thanks @kunalq for picking this up!

95-martin-orion (approved):

PR #369, opened 2021-06-23
Update version to 0.10.0.dev20210623

95-martin-orion:
CC @kunalq 

This is required for the dev release fix for #349. The [Python docs](https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning) indicate that this will be appropriately treated as a pre-release branch.


kunalq (approved):
Thank you!

sergeisakov (approved):

PR #371, opened 2021-06-27
MPS Implementation p3: Added ToWaveFunction helper.

MichaelBroughton:
Adds the `ToWaveFunction` testing method to make MPS feature development and verification easier. Also aliased `std::complex<fp_type>` to `Complex` to make things a little more readable.


tests/mps_statespace_test.cc
```diff
@@ -75,11 +75,304 @@ TEST(MPSStateSpaceTest, Copy) {
...
+    EXPECT_NEAR(wf[i], 0, 1e-5);
+  }
+  delete[](wf);
+}
+
+TEST(MPSStateSpaceTest, ToWaveFunction3) {
+  auto ss = MPSStateSpace<For, float>(1);
+  auto mps = ss.CreateMPS(3, 4);
+
+  // Set to highly entangled three qubit state.
```
95-martin-orion, line 95: What was used to contruct these MPS states and wavefunctions? If an error occurred in this test, I wouldn't know what the issue was outside of "the MPS simulator is broken."

lib/mps_statespace.h
```diff
@@ -144,6 +155,37 @@ class MPSStateSpace {
...
+    ConstMatrixMap accum = ConstMatrixMap((Complex*)(raw_state), 2, bd);
+    ConstMatrixMap next_block = ConstMatrixMap(nullptr, 0, 0);
+    MatrixMap result2 = MatrixMap(nullptr, 0, 0);
+    auto offset = 0;
+    auto result2_size = 2;
+
+    for (unsigned i = 1; i < nq - 1; i++) {
+      offset = GetBlockOffset(state, i);
+      // use of new does not trigger any expensive operations.
+      new (&next_block)
```
95-martin-orion, line 174: I'm not familiar with this usage of `new` - is it available in C++11? (This is required for use with pybind)

I assume this is using `next_block` to store the `ConstMatrixMap` constructed here. Does this behave similarly to something like this?
```
ConstMatrixMap* next_block = new ConstMatrixMap(nullptr, 0, 0);
*next_block = ConstMatrixMap(...);
```
(apologies for errors, I've been away from C++ for a while)

lib/mps_statespace.h
```diff
@@ -144,6 +155,37 @@ class MPSStateSpace {
...
+
+      // temp variable used since result2 and accum point to same memory.
+      result2 = accum * next_block;
+      result2_size *= 2;
+      new (&accum) ConstMatrixMap((Complex*)(wf), result2_size, bd);
+    }
+    offset = GetBlockOffset(state, nq - 1);
+    new (&next_block) ConstMatrixMap((Complex*)(raw_state + offset), bd, 2);
+    new (&result2) MatrixMap((Complex*)(wf), result2_size, 2);
+    result2 = accum * next_block;
```
95-martin-orion, line 186: Is this somehow mutating the contents of `state` or `wf`?

95-martin-orion (changes requested):
Couple of questions and comments, mostly driven by my unfamiliarity with certain C++ syntax.

tests/mps_statespace_test.cc
MichaelBroughton: It's just an arbitrary wavefunction I made using numpy. From there I worked backwards to MPS matrix entries. You could also go the other way and produce some random matrix entries and then infer what the state vector ought to be. Do you have any suggestions on improving these tests ?

lib/mps_statespace.h
MichaelBroughton: Thematically it does do something like the snippet you've shown. See the "Changing Mapped Array" section: https://eigen.tuxfamily.org/dox/group__TutorialMapClass.html . It is available in C++ 11.

lib/mps_statespace.h
MichaelBroughton: `result2` is a `MatrixMap` which means it can be written to (unlike `ConstMatrixMap`). This matrix multiplication is modifying the contents of `wf`. It is not modifying any values in `raw_state`.

tests/mps_statespace_test.cc
95-martin-orion: I'd be more confident in a test that verified this against some other MPS-to-wavefunction conversion, or provided comments detailing how to do that conversion. In the end, though, this is probably more my own lack of familiarity showing rather than a true issue with the test - I'll leave it up to you whether to clarify further.

lib/mps_statespace.h
95-martin-orion: TIL "placement new" is a thing. Thanks!

95-martin-orion (approved):
No further issues from me, aside from the remaining (optional) clarification.

tests/mps_statespace_test.cc
MichaelBroughton: I personally think this isolates the behavior of the wavefunction conversion reasonably well. If we relied on some other MPS to wavefunction converter to check our own MPS to wavefunction converter it kind of feels like the unit test wouldn't be very targeted and would be more like some kind of fuzz test. I'll go ahead and add a comment explaining the desired transformation we want the MPS blocks to undergo for each test.

95-martin-orion: Thanks, the comment makes this much clearer!

Issue #361, opened 2021-06-10
Detect SIMD architecture from qsimcirq


95-martin-orion:
Related to #313. In order for users to have access to the instruction set their machine supports, we need to allow qsimcirq to dynamically choose between the different simulator implementations present in C++. Currently this is handled by the use of `simmux.h` in the pybind layer, but to enable the use of wheels additional handling may be required.

sergeisakov:
One can add a template parameter to `SimulatorHelper` and modify functions like `qsim_simulate_expectation_values`. That is, for instance,

```C++
std::vector<std::complex<double>> qsim_simulate_expectation_values(
    const py::dict &options,
    const std::vector<std::tuple<
                          std::vector<OpString<Cirq::GateCirq<float>>>,
                          unsigned>>& opsums_and_qubit_counts,
    uint64_t input_state) {
  return SimulatorHelper::simulate_expectation_values(
    options, opsums_and_qubit_counts, false, input_state);
}
```

can be modified as follows

```C++
std::vector<std::complex<double>> qsim_simulate_expectation_values(
    const py::dict &options,
    const std::vector<std::tuple<
                          std::vector<OpString<Cirq::GateCirq<float>>>,
                          unsigned>>& opsums_and_qubit_counts,
    uint64_t input_state) {
  auto instruction_set = DetectInstructions();

  switch (instruction_set) {
  case kAVX512:
    return SimulatorHelper<SimulatorAVX512<For>>::simulate_expectation_values(
      options, opsums_and_qubit_counts, false, input_state);
  case kAVX:
    return SimulatorHelper<SimulatorAVX<For>>::simulate_expectation_values(
      options, opsums_and_qubit_counts, false, input_state);
  case kSSE:
    return SimulatorHelper<SimulatorSSE<For>>::simulate_expectation_values(
      options, opsums_and_qubit_counts, false, input_state);
  default:
    return SimulatorHelper<SimulatorBasic<For>>::simulate_expectation_values(
      options, opsums_and_qubit_counts, false, input_state);
  }
}
```

We need to implement `DetectInstructions` that detects the SIMD instruction set. Also some additional changes are needed because not every case is covered by `SimulatorHelper`.

95-martin-orion:
I think we can fulfill #334 with this by allowing users to pass in `instruction_set` to these methods (defaulting to `DetectInstructions` if not specified).

[...]

PR #378, opened 2021-07-14
Separate simd compilation units

laurynasas:
This is work targeting issue in https://github.com/quantumlib/qsim/issues/361.


95-martin-orion (commented):
This approach seems alright to me, given that the majority of tests are still passing. The Docker failure seems like it might be a linking issue? Since now `pybind_main.cpp` only makes sense in the context of one of the `pybind_main_*` files.

laurynasas:
right, that's correct, fixed this, PTAL

95-martin-orion (approved):

PR #379, opened 2021-07-19
Wheels macos

laurynasas:
This PR addresses a cross-platform wheels [issue](https://github.com/quantumlib/qsim/issues/313) and adds support for building and uploading MacOS wheels for cp36, cp37, cp38, cp39 on x86_64 architecture.


95-martin-orion (commented):
It looks like this PR only handles the release workflow. Is the intent to rely on the existing Kokoro pipeline for verifying the MacOS build on pull requests?

Given that we'll need Kokoro for testing GPU and AVX512 builds regardless, this would be a valid choice - just want to confirm.

laurynasas:
@95-martin-orion I added two separate workflows for building the wheels, one for releases with the upload to the PyPI and the other for PRs and merges to master for testing purposes. Is that what you expected?

95-martin-orion (approved):
Yes, this looks good to me - and it looks like it will run as part of this PR, so we'll have clear confirmation that it works.

Issue #340, opened 2021-05-05
Reduce circuit translation overhead


95-martin-orion:
Memoizing Cirq-to-qsim circuit translations in the QSimSimulator should reduce the overhead cost of circuit translation.

A more complete solution requires Cirq-level redesign of the simulator API, including [Cirq #4064](https://github.com/quantumlib/Cirq/issues/4064) to prevent results from multiple repetitions from being stored in memory simultaneously.

verult:
I'm happy to take a look at this.

verult:
For my use case of running trajectory simulations with multiple repetitions, memoizing only the last circuit is sufficient. Are there other use cases where a larger data structure is necessary?

95-martin-orion:
In the near term, one possible use case for multiple-circuit memoization is "chunked circuits". Since qsim does not yet support intermediate state simulation, if a user wants to get the state partway through the circuit they must simulate the "chunk" before that point, get the result, then use that result as the input for the next "chunk".

Repeated simulation of such a circuit would benefit from memoizing all "chunks", which may each be a distinct circuit.

verult:
We could potentially add a new constructor parameter `circuit_memoization_size` to indicate how many chunks should be kept.

Looks like `Circuit`s are not hashable (presumably because it's mutable?). Are there other hashing strategies offered by Cirq, or do we have to resort to a linear search?

95-martin-orion:
There are a couple of options for hashing, with considerations for each:
- Circuits can be "frozen" into a hashable form with `circuit.freeze()`. Equality comparison of this frozen format takes time proportional to the size of the circuit (i.e. number of gates)
- The Python id of the circuit can be used for hashing (e.g. `id(circuit)`). Comparison time for this is constant, but copies of a circuit won't have the same id even if they have identical gates.

verult:
Does the FrozenCircuit happen to guarantee not having hash collisions for different circuits, i.e. can we bypass the equality check?

95-martin-orion:
> Does the FrozenCircuit happen to guarantee not having hash collisions for different circuits, i.e. can we bypass the equality check?

The larger issue here is that `hash(FrozenCircuit)` is not a constant-time operation - it's O(N) for an N-gate circuit.

As far as hashing goes, I'm hesitant to provide a "no-collisions" guarantee as there was staunch opposition when I attempted to use a similar guarantee in FrozenCircuit serialization.

verult:
[...]

PR #387, opened 2021-07-29
QSimSimulator: memoizing circuit translations

verult:
Fixes https://github.com/quantumlib/qsim/issues/340

This implementation uses a queue as the memoization data structure, and circuit match is based on a linear search through the queue using circuit equality checks.

For simplicity, when there is a cache hit, the circuit does not move to the head of the queue. This is fine for the use cases mentioned in https://github.com/quantumlib/qsim/issues/340.

@95-martin-orion

### DIFF
--- a/qsimcirq_tests/qsimcirq_test.py
+++ b/qsimcirq_tests/qsimcirq_test.py
@@ -1376,3 +1376,152 @@ def test_cirq_qsim_global_shift():
     assert cirq.linalg.allclose_up_to_global_phase(
         qsim_result.state_vector(), cirq_result.state_vector()
     )
+
+
+@pytest.mark.parametrize("mode", ["noiseless", "noisy"])
+def test_cirq_qsim_circuit_memoization_compute_amplitudes(mode: str):
+    """Verifies the correctness of simulator functions when
+    circuit_memoization_size is set."""
+    execution_repetitions = 3
+    qsim_sim = qsimcirq.QSimSimulator(circuit_memoization_size=4)
+
+    # Pick qubits.
+    a, b, c, d = [
+        cirq.GridQubit(0, 0),
+        cirq.GridQubit(0, 1),
+        cirq.GridQubit(1, 1),
+        cirq.GridQubit(1, 0),
+    ]
+
+    # Create a circuit
+    cirq_circuit = cirq.Circuit(
+        cirq.X(a) ** 0.5,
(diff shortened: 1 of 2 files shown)


qsimcirq_tests/qsimcirq_test.py
```diff
@@ -1376,3 +1376,123 @@ def test_cirq_qsim_global_shift():
...
+
+    # Pick qubits.
+    a, b, c, d = [
+        cirq.GridQubit(0, 0),
+        cirq.GridQubit(0, 1),
+        cirq.GridQubit(1, 1),
+        cirq.GridQubit(1, 0),
+    ]
+
+    # compute_amplitudes()
```
95-martin-orion, line 1397: Could you split these (`compute_amplitudes`, `simulate`, etc.) into separate test methods?

95-martin-orion (changes requested):
Looks great overall! Do you have an example of how much this improves performance for many repetitions of (relatively) small circuits?

verult:
Basing off of the cirq quantum phase estimation example, I benchmarked with the code below:

```python
import cirq
import qsimcirq
import timeit
import numpy as np


def run():
    repetitions = 1000
    qnum = 8
    phi = 0.2
    gate = cirq.MatrixGate(matrix=np.array([[np.exp(2 * np.pi * 1.0j * phi), 0], [0, 1]]))

    ancilla = cirq.LineQubit(-1)
    qubits = cirq.LineQubit.range(qnum)

    oracle_raised_to_power = [
        gate.on(ancilla).controlled_by(qubits[i]) ** (2 ** i) for i in range(qnum)
    ]
    circuit = cirq.Circuit(
        cirq.H.on_each(*qubits),
        oracle_raised_to_power,
        cirq.qft(*qubits, without_reverse=True) ** -1,
    )

    simulator = qsimcirq.QSimSimulator(circuit_memoization_size=1)
    #simulator = qsimcirq.QSimSimulator()
    for _ in range(repetitions):
        simulator.simulate(circuit)


print("Execution time: ", timeit.timeit(run, number=1))
```

On my machine: without memoization: ~3.7s, with memoization: ~0.6s.

qsimcirq_tests/qsimcirq_test.py
verult: done :)

95-martin-orion (approved):

verult:
addressed python format failure

PR #389, opened 2021-07-30
Wheels win

laurynasas:


sergeisakov:
Applications in `apps` are sample applications for experts. They were never meant to run on Windows. So I suggest that we don't add support for Windows in `apps`.

laurynasas:
@sergeisakov Thanks for letting me know, I removed the Windows support from the `apps`

95-martin-orion (approved):

PR #392, opened 2021-08-02
Add qtrajectory CUDA tests.

sergeisakov:


tests/BUILD
```diff
@@ -173,24 +173,38 @@ cc_test(
...
         "//lib:gates_cirq",
         "//lib:io",
         "//lib:qtrajectory",
-        "//lib:simulator",
+    ],
+    testonly = 1,
+)
+
+cc_test(
+    name = "qtrajectory_avx_test",
```
95-martin-orion, line 197: Do you see any need for SSE/basic versions of this test?

95-martin-orion (approved):
Great to see how clean this change is! Just one comment regarding test coverage.

tests/BUILD
sergeisakov: No, I don't really see any need for SSE/basic versions. It doesn't add any additional test coverage.

PR #394, opened 2021-08-03
Fix MPS import in TFQ

### LINKED ISSUE #393: mps is not included in BUILD.

jaeyoo:
Resolves #393


lib/BUILD
```diff
@@ -125,6 +129,8 @@ cc_library(
         "io.h",
         "io_file.h",
         "matrix.h",
+        "mps_simulator.h",
+        "mps_statespace.h",
```
95-martin-orion, line 133: qsimh (the hybrid Schrodinger-Feynman simulator) is largely unused at this point. Is it necessary to include these here?

95-martin-orion (changes requested):
Mostly seems fine, one comment on qsimh.

lib/BUILD
jaeyoo: Oh my mistake. let me remove it then :) Thank you @95-martin-orion

95-martin-orion (approved):
Looks good to me. To confirm, #393 is resolved by this, but not #395 - is that correct?

jaeyoo:
> Looks good to me. To confirm, #393 is resolved by this, but not #395 - is that correct?

Yes, right. #394 resolves #393 . #396 resolves #395 . Thanks!

sergeisakov:
> Mostly seems fine, one comment on qsimh.

I think the same comment is valid for "run_qsim_lib". This should have been added only to "qsim_lib".

Issue #380, opened 2021-07-20
Opt in to Github Actions MacOS 11.0 VM support


laurynasas:
Please see the [doc](https://github.com/actions/virtual-environments/blob/main/docs/macos-11-onboarding.md) for more information on how a project can get enrolled for an exclusive access to macos-11.0 VMs in Github Actions.

laurynasas:
Filled in the form, will wait for confirmation.

mhucka:
Closing because this looks like it was done at the time (and anyway, these days it's been superceded by later Mac versions).

PR #397, opened 2021-08-05
add macos11 wheels

laurynasas:
Addresses issue mentioned in https://github.com/quantumlib/qsim/issues/380

### DIFF
--- a/.github/workflows/release_wheels.yml
+++ b/.github/workflows/release_wheels.yml
@@ -12,12 +12,21 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - os: macos-11
+            name: mac11
+            cibw:
+              build: "cp36* cp37* cp38*"
+          - os: macos-11
+            name: mac11-arm
+            cibw:
+              arch: universal2
+              build: "cp39*"
           - os: macos-10.15
-            name: mac
+            name: mac10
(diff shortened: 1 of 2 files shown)


laurynasas:
This is actually not needed as cibuildwheel MacOS wheel supports MacOS 10.9+ https://github.com/pypa/cibuildwheel#what-does-it-do

95-martin-orion:
> This is actually not needed as cibuildwheel MacOS wheel supports MacOS 10.9+ https://github.com/pypa/cibuildwheel#what-does-it-do

Do we have a way to confirm that this means (10.9)+ and not 10.(9+)?

laurynasas:
> > This is actually not needed as cibuildwheel MacOS wheel supports MacOS 10.9+ https://github.com/pypa/cibuildwheel#what-does-it-do
> 
> Do we have a way to confirm that this means (10.9)+ and not 10.(9+)?

I attempted that locally on my Big Sur machine by successfully installing 10.9 wheel and running the qsimcirq tests. Should we look into automated tests on macos-11 machines? Any other suggestions?

95-martin-orion:
As long as you were able to confirm it locally, I think the existing test cases will be sufficient. Thanks for verifying!

PR #402, opened 2021-08-06
Add CUDA hybrid test.

sergeisakov:


tests/hybrid_cuda_test.cu
```diff
@@ -0,0 +1,66 @@
...
+
+  Simulator CreateSimulator() const {
+    return Simulator(param2);
+  }
+
+  typename StateSpace::Parameter param1;
+  typename Simulator::Parameter param2;
+};
+
+TEST(HybridAVXTest, Hybrid2) {
```
95-martin-orion, line 45: `HybridAVXTest` -> `HybridCUDATest`

Same below.

95-martin-orion (approved):
Tests are passing - just one naming fix and this is good to go.

tests/hybrid_cuda_test.cu
sergeisakov: Fixed.

PR #413, opened 2021-08-16
[WIP] Save time in Docker tests

### LINKED ISSUE #411: Increased runtime for Docker tests
Starting with #379 (merged on July 26), Docker tests began taking ~10 minutes longer to complete on average. Either that PR or some other change around that time is suspected to have caused this. We should investigate to see if we can bring the runtime back down, as this is starting to negatively affect the pace at which we can merge PRs.

95-martin-orion:
Fixes #411.

This PR is a catch-all for changes to improve the runtime of Docker tests.

### DIFF
--- a/pybind_interface/Dockerfile
+++ b/pybind_interface/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get install -y python3-dev python3-pybind11 python3-pytest python3-pip
 
 # The --force flag is used mainly so that the old numpy installation from pybind
 # gets replaced with the one cirq requires
-RUN pip3 install cirq-core --force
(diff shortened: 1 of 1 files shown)


95-martin-orion:
Installing `cirq-core` instead of `cirq` improved runtime slightly: no `grpcio` wheel means saves ~5 minutes.

95-martin-orion:
Adding the `--prefer-binary` flag causes Docker tests to no longer be the bottleneck.

sergeisakov (approved):
Thank you. It seems "Build wheels on windows-2019" is the bottleneck now.

PR #417, opened 2021-08-19
Patch fixes for importing qsim to Google

95-martin-orion:
These changes are necessary to appease the import process:
- Tagging `cc_library` rules with CUDA dependencies allows us to convert them to `cuda_library` rules internally.
- Removing `auto sim =` in the test silences an unused variable warning which internal tests treat as an error.

### DIFF
--- a/lib/BUILD
+++ b/lib/BUILD
@@ -1,12 +1,12 @@
 package(default_visibility = ["//visibility:public"])
 
-##### Aggregate libraries #####
-
 # Libraries of the following form:
 #   # cuda_library
 #   cc_library(...)
 # are converted to cuda_library rules when imported to the Google codebase.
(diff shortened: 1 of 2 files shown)


lib/BUILD
```diff
@@ -2,7 +2,14 @@ package(default_visibility = ["//visibility:public"])
 
 ##### Aggregate libraries #####
 
+# Libraries of the following form:
```
sergeisakov, line 5: Perhaps move this comment above `##### Aggregate libraries #####` as this tag is also used with other libraries?

sergeisakov (approved):

lib/BUILD
95-martin-orion: Good point - done.

Issue #418, opened 2021-08-19
Remove libomp and llvm installation requirement from MacOS wheels


laurynasas:
Currently to install MacOS wheel user needs to install `libomp` and `llvm`. This can be fixed by "repairing" the wheel with delocate and including the dependencies to the wheel.

mhucka:
I believe this was address in a past PR (#419) but this issues was never closed at the time. Closing now.

PR #419, opened 2021-08-19
enable delocate to fix the macos wheels

laurynasas:
addresses https://github.com/quantumlib/qsim/issues/418


95-martin-orion (approved):
A very belated approval on this to fix #535. Thanks @laurynasas for your foresight!

PR #420, opened 2021-08-20
Split qsim_lib into CUDA/CC versions

95-martin-orion:
The `cuda_library` BUILD rule is "contagious" - any `cc_library` rule that would depend on it must be converted to a `cuda_library` rule. To avoid this (it causes issues internally), I've split `qsim_lib` into C++ and CUDA variants.

`qsim_lib` isn't referenced anywhere else in the qsim repo, so I don't expect issues from this change.

### DIFF
--- a/lib/BUILD
+++ b/lib/BUILD
@@ -8,10 +8,66 @@ package(default_visibility = ["//visibility:public"])
 
 ##### Aggregate libraries #####
 
-# Full qsim library
-# cuda_library
+# Full qsim library, minus CUDA
 cc_library(
     name = "qsim_lib",
+    hdrs = [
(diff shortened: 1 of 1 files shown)


sergeisakov (approved):

sergeisakov:
> `qsim_lib` isn't referenced anywhere else in the qsim repo, so I don't expect issues from this change.

What about TFQ?

95-martin-orion:
> What about TFQ?

TFQ currently depends on `qsim_lib`, but does not yet depend on the CUDA support it provides. Once they want CUDA support, they can migrate to `qsim_cuda_lib`.

PR #442, opened 2021-10-01
Fix GPU install process

95-martin-orion:
The previous cmake process didn't properly install `qsimcirq.qsim_gpu`, which meant users had to rely on `make` and only work inside the qsim directory. This PR updates the cmake to support installing GPU.

Tested on:

- [x] Linux, no GPU
- [x] Linux with GPU
- [x] Windows, no GPU
- [x] MacOS, no GPU

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,9 @@
 cmake_minimum_required(VERSION 3.11)
-project(qsim)
+project(qsim LANGUAGES CXX CUDA)
 
 ADD_SUBDIRECTORY(pybind_interface/sse)
 ADD_SUBDIRECTORY(pybind_interface/avx512)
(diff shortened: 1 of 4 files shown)


95-martin-orion:
As far as I can tell, the CI "build wheel" tests also invoke setup.py and CMake. Since they're passing, I am pleasantly surprised to report that this PR works for the (non-GPU) MacOS and Windows build paths.

95-martin-orion:
@cognigami It should be possible to test this with the "GPU on GCP" workflow by switching to the `qsim-gpu-gcp` branch after cloning the repository.

sergeisakov (approved):
LGTM

95-martin-orion:
Kokoro build appears to be a flake (timed out trying to download `com_google_test`). Docker failure is related to #444.

PR #467, opened 2021-10-30
refactor: split dev requirements out of main requirements.txt

lovesegfault:
Currently, if you try to use a tool like [Poetry][1] to specify a dependency on
qsim you end up with the unexpected behavior of your package's version of
`black` having to be pinned to same version `qsim` pins in its `requirements.txt`.

This should solve this by splitting development dependencies out of the main dependency file.

[1]: https://python-poetry.org

### DIFF
--- a/.github/workflows/python_format.yml
+++ b/.github/workflows/python_format.yml
@@ -24,9 +24,7 @@ jobs:
         with:
           python-version: '3.7'
           architecture: 'x64'
-      - name: Install flynt
-        run: cat requirements.txt | grep flynt | xargs pip install
-      - name: Install black
-        run: cat requirements.txt | grep black | xargs pip install
+      - name: Install dev requirements
+        run: pip install -r dev-requirements.txt
       - name: Format
         run: check/format-incremental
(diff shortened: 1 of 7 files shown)


95-martin-orion:
This change will likely also require updates to existing documentation to ensure that developers know to install the additional requirements file manually, and to the CI workflows so that tests will pass.

lovesegfault:
> This change will likely also require updates to existing documentation to ensure that developers know to install the additional requirements file manually, and to the CI workflows so that tests will pass.

Done! I think this is ready for review now :)

95-martin-orion:
Code changes all LGTM. Do you know if this will play nicely with our existing notebooks, though? I have a suspicion that moving the `pybind11` dependency to a separate package might break things when pip-installing.

lovesegfault:
> Code changes all LGTM. Do you know if this will play nicely with our existing notebooks, though? I have a suspicion that moving the `pybind11` dependency to a separate package might break things when pip-installing.

People will have to do `pip install qsimcirq[dev]` now to get all the deps. Happy to test this if you can show me how.

95-martin-orion:
Testing is somewhat inconvenient, given that it requires a qsimcirq release (preferably on test-pypi). Instead, if `black` and `flynt` are the main concerns here, could you shift `pybind11` from the dev requirements back into the main requirements? That should resolve my concerns so tests aren't necessary.

lovesegfault:
> Testing is somewhat inconvenient, given that it requires a qsimcirq release (preferably on test-pypi). Instead, if `black` and `flynt` are the main concerns here, could you shift `pybind11` from the dev requirements back into the main requirements? That should resolve my concerns so tests aren't necessary.

Done!

95-martin-orion (approved):

PR #479, opened 2021-11-16
[WIP] Include GPU in wheels

### LINKED ISSUE #430: Provide GPU support in released qsimcirq wheel
Currently, qsimcirq users who want to run on GPU must compile the code locally, as the wheels are built on Github Actions runners which [do not have the CUDA SDK](https://github.com/actions/virtual-environments/issues/116). Compared to the CPU workflow of `pip install qsimcirq`, this is incovenient; it would be preferable if GPU support was built into the released wheels.

In order to provide this support, we need to investigate the following options (listed from most to least preferable):

1. Create an image with the CUDA SDK and use it in the release.
    - Minimizes the impact to our release speed
    - Still need to determine if this is possible in Github Actions 
2. Install the CUDA SDK on the Github Actions runner during each release.
    - Definitely possible in Github Actions (an action exists for installing the SDK)
    - May increase release (not CI) runtime by 10+ minutes
3. Establish a secondary workflow for releasing the GPU wheels.
    - Allows us to use our own devices with the CUDA SDK pre-installed
    - Increases maintenance costs, and a single wheel would be preferable

95-martin-orion:
Fixes #430.

Since cross-platform support requires building on the manylinux image, installing NVCC on the runner is the best we can do unless we decide to push a manylinux + CUDA image to Docker Hub.

### DIFF
--- a/.github/workflows/testing_wheels.yml
+++ b/.github/workflows/testing_wheels.yml
@@ -52,6 +52,11 @@ jobs:
       # Used to host cibuildwheel
       - uses: actions/setup-python@v2
 
+      # Install NVCC on Ubuntu to include GPU in the wheel.
+      - name: cuda-toolkit
+        if: ${{ matrix.os == 'ubuntu-20.04' }}
+        uses: Jimver/cuda-toolkit@v0.2.4
+
       - name: Install cibuildwheel and twine
         run: python -m pip install cibuildwheel==1.11.0
 


95-martin-orion:
Bazel page is returning 404s, need to wait for it to come back and retry.

Nexuscompute (approved):

95-martin-orion:
An inevitable side effect of building GPU support on machines that lack GPU support: we have no way to test GPU behavior in continuous integration. However, since the wheels generated by GHA are accessible from the test artifacts, it will be possible for us to spin up a GPU-enabled machine, copy over the wheels, and test before release.

If we move forward with this, the process outlined above will become a mandatory part of all GPU changes.

leofang:
@95-martin-orion @sergeisakov FYI, we prepared a demo to show how to build a project upon CUDA/cuQuantum wheels: https://github.com/NVIDIA/cuQuantum/tree/main/extra/demo_build_with_wheels
Let us know if it helps!

PR #483, opened 2021-11-19
Import and use cirq instead of submodules

maffoo:
Since qsimcirq is external to cirq, we can refer to names from the top-level `cirq` namespace.

### DIFF
--- a/qsimcirq/qsim_simulator.py
+++ b/qsimcirq/qsim_simulator.py
@@ -214,7 +214,7 @@ class QSimSimulator(
                 'Keys {"c", "i", "s"} are reserved for internal use and cannot be '
                 "used in QSimCircuit instantiation."
             )
-        self._prng = cirq.parse_random_state(seed)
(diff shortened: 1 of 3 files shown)


qsimcirq/qsim_simulator.py
```diff
@@ -16,18 +16,7 @@
...
-    study,
-    value,
-    SimulatesAmplitudes,
-    SimulatesFinalState,
-    SimulatesSamples,
-)
+import cirq
 
 # TODO: import from cirq directly when fix is released
 from cirq.sim.simulator import SimulatesExpectationValues
```
95-martin-orion, line 22: This TODO is resolved now - `cirq.SimulatesExpectationValues` can replace this.

95-martin-orion (approved):
One request, if you don't mind adding it. The rest LGTM

qsimcirq/qsim_simulator.py
maffoo: Done.

PR #484, opened 2021-11-23
Support invert_mask in qsimcirq

### LINKED ISSUE #482: QSimSimulator does not respect invert_mask on measurements
Simple example:
```
In [1]: import cirq, qsimcirq

In [2]: circuit = cirq.Circuit(cirq.measure(cirq.LineQubit(0), key='a', invert_mask=[True]))

In [3]: cirq.Simulator().sample(circuit, repetitions=5)
Out[3]: 
   a
0  1
1  1
2  1
3  1
4  1

In [4]: qsimcirq.QSimSimulator().sample(circuit, repetitions=5)
Out[4]: 
   a
0  0
1  0
2  0
3  0
4  0
```

95-martin-orion:
Fixes #482.


95-martin-orion:
This is a first pass at this - the tests are almost certainly not comprehensive (still need to verify different `qubit_order`s, measuring on different qubits, etc.)

95-martin-orion:
Turns out `sample` doesn't take a `qubit_order` argument, so this may be sufficient. LMK if there are any other tests you'd like to see for this.

qsimcirq/qsim_simulator.py
```diff
@@ -391,8 +392,11 @@ def _sample_measure_results(
                 options["s"] = self.get_seed()
                 measurements = sampler_fn(options)
                 for key, bound in bounds.items():
+                    invert_mask = meas_ops[key].gate.full_invert_mask()
                     for j in range(bound[1] - bound[0]):
-                        results[key][i][j] = int(measurements[bound[0] + j])
+                        results[key][i][j] = int(
+                            measurements[bound[0] + j] ^ invert_mask[j]
+                        )
```
maffoo, line 399: Drive-by nit: doing this 3d loop in python, especially over the repetitions axis which could be very large, seems very inefficient. It'd be nice to change this to use numpy array operations to assign things in bulk as much as possible.

maffoo (approved):

qsimcirq/qsim_simulator.py
95-martin-orion: Thanks for highlighting this - I'll open a follow-up PR to improve on it.

PR #485, opened 2021-11-30
Optimize invert mask application

95-martin-orion:
Adds numpy optimization in response to comments on #484.

In the second (noisy) block, this requires additional memory to store measurement results for all repetitions simultaneously. However, (repetitions * measurements) boolean values should be trivially small compared to the state vectors used in simulation.

In theory the existing tests cover this behavior, but if you have a particular use case in mind I'd appreciate if you give it a try with this before we merge. I _think_ this behaves the same as before based on the documentation, but my numpy-fu is weak.


qsimcirq/qsim_simulator.py
```diff
@@ -376,27 +376,34 @@ def _sample_measure_results(
...
-                    meas_indices = [qubit_map[qubit] for qubit in op.qubits]
-                    invert_mask = op.gate.full_invert_mask()
-                    for j, q in enumerate(meas_indices):
-                        results[key][i][j] = full_results[i][q] ^ invert_mask[j]
+            for key, op in meas_ops.items():
+                meas_indices = [qubit_map[qubit] for qubit in op.qubits]
+                invert_mask = op.gate.full_invert_mask()
+                # Match result order to order in ops, then apply invert mask
+                permuted_results = full_results[:, meas_indices]
+                results[key] = np.logical_xor(permuted_results, invert_mask)
```
maffoo, line 384: I think the `^` operator should work fine on arrays:
```suggestion
                results[key] = permuted_results ^ invert_mask
```

qsimcirq/qsim_simulator.py
```diff
@@ -376,27 +376,34 @@ def _sample_measure_results(
...
-                    for j in range(bound[1] - bound[0]):
-                        results[key][i][j] = int(
-                            measurements[bound[0] + j] ^ invert_mask[j]
-                        )
+                measurements[i] = sampler_fn(options)
+
+            for key, bound in bounds.items():
+                invert_mask = meas_ops[key].gate.full_invert_mask()
+                permutation = list(range(bound[0], bound[1]))
+                permuted_measurements = measurements[:, permutation]
```
maffoo, line 405: Could destructure the `bound` tuple in the for loop, and use a slice to extract the needed values:
```python
for key, (start, end) in bounds.items():
    invert_mask = meas_ops[key].gate.full_invert_mask()
    permuted_measurements = measurements[:, start:end]
```

qsimcirq/qsim_simulator.py
```diff
@@ -376,27 +376,34 @@ def _sample_measure_results(
...
+                results[key] = np.logical_xor(permuted_measurements, invert_mask)
```
maffoo, line 406: Here, too, the `^` operator should work on arrays:
```suggestion
                results[key] = permuted_measurements ^ invert_mask
```
Also, I think "permuted` is a bit misleading since we don't change the order, just extract a slice of qubits with the same order, so perhaps could just inline the slice expression here to avoid a potentially misleading name:
```suggestion
                results[key] = measurements[:, start:end] ^ invert_mask
```

maffoo (approved):
Minor comments, then LGTM.

qsimcirq/qsim_simulator.py
95-martin-orion: Got `TypeError: ufunc 'bitwise_xor' not supported for the input types` - maybe because `invert_mask` isn't an `np.array`? Sticking with `np.logical_xor` for now.

qsimcirq/qsim_simulator.py
95-martin-orion: Done.

qsimcirq/qsim_simulator.py
95-martin-orion: Applied the slice but kept `np.logical_xor`, same as above.

qsimcirq/qsim_simulator.py
maffoo: Hm, we use `^` internally, so I'm wondering what the difference is here. What is the dtype of the `measurements` array? Also, it looks like `np.logical_xor` coerces the output to be a bool array, while `^` (which calls `np.bitwise_xor` under the hood) does not:
```
In [11]: np.zeros((3, 2), dtype=np.uint8) ^ [True, False]
Out[11]: 
array([[1, 0],
       [1, 0],
       [1, 0]], dtype=uint8)

In [12]: np.logical_xor(np.zeros((3, 2), dtype=np.uint8), [True, False])
Out[12]: 
array([[ True, False],
       [ True, False],
       [ True, False]])
```

95-martin-orion: Good call on the `dtype` - it was defaulting to `float64` which didn't want to XOR with booleans.

95-martin-orion:
Resolved the `logical_xor` vs `^` issue - this is ready to merge.

maffoo (approved):

PR #496, opened 2022-01-05
immintrin.h is not included in CUDA code.

sergeisakov:


95-martin-orion (approved):

95-martin-orion:
Confirmed that this unblocked #479 - it's now facing test failures, but those are far more approachable.

PR #499, opened 2022-01-14
Intermediate expectation values.

95-martin-orion:
Adds `simulate_moment_expectation_values` to the QSimSimulator API.

This method accepts a dict of (moment index: observable list), and calculates the requested observables at the specified moments. Special-case behavior is provided for calculating the same observable (or list of observables) at every moment.

### DIFF
--- a/pybind_interface/pybind_main.cpp
+++ b/pybind_interface/pybind_main.cpp
@@ -594,7 +594,7 @@ class SimulatorHelper {
     // Aggregate expectation values for noisy circuits.
     for (unsigned i = 0; i < opsums_and_qubit_counts.size(); ++i) {
       auto& counts = std::get<1>(opsums_and_qubit_counts[i]);
-      results[i] = std::vector<std::complex<double>>(counts.size(), 0);
+      results[i].resize(counts.size(), 0);
     }
     for (unsigned rep = 0; rep < helper.noisy_reps; ++rep) {
       // Init outside of simulation to enable stepping.
--- a/pybind_interface/pybind_main.cpp
+++ b/pybind_interface/pybind_main.cpp
@@ -729,18 +729,15 @@ class SimulatorHelper {
     if (is_noisy) {
       std::vector<uint64_t> stat;
       auto params = get_noisy_params();
-      NoisyCircuit<Gate> subncircuit;
-      subncircuit.num_qubits = ncircuit.num_qubits;
-      subncircuit.channels = std::vector<Channel<Gate>>(
-        ncircuit.channels.begin() + begin,
-        ncircuit.channels.begin() + end
-      );
-
       Simulator simulator = factory.CreateSimulator();
       StateSpace state_space = factory.CreateStateSpace();
 
-      result = NoisyRunner::RunOnce(params, subncircuit, seed, state_space,
-                                    simulator, scratch, state, stat);
+      result = NoisyRunner::RunOnce(
+        params, ncircuit.num_qubits,
+        ncircuit.channels.begin() + begin,
+        ncircuit.channels.begin() + end,
+        seed, state_space, simulator, scratch, state, stat
+      );
     } else {
       Circuit<Gate> subcircuit;
       subcircuit.num_qubits = circuit.num_qubits;
--- a/pybind_interface/pybind_main.cpp
+++ b/pybind_interface/pybind_main.cpp
@@ -273,6 +273,25 @@ void control_last_gate(const std::vector<unsigned>& qubits,
   MakeControlledGate(qubits, values, circuit->gates.back());
 }
 
+void compose_circuits(Circuit<Cirq::GateCirq<float>>* target,
+                      Circuit<Cirq::GateCirq<float>>* suffix) {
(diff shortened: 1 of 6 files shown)


95-martin-orion:
I have tested this locally, but have not attempted it on a GPU device.

qsimcirq/qsim_simulator.py
```diff
@@ -718,3 +857,28 @@ def _translate_circuit(
             self._translated_circuits.append((circuit, translated_circuit))
 
         return translated_circuit
+
+    def _translate_moments(
```
sergeisakov, line 861: There is no circuit memoization, right?

qsimcirq/qsim_simulator.py
```diff
@@ -718,3 +857,28 @@ def _translate_circuit(
...
+        reindex = [0]
+        for moment in circuit:
+            subc = self._translate_circuit(
+                qsimc.QSimCircuit(cirq.Circuit(moment)),
+                translator_fn_name,
+                qubit_order,
+            )
+            reindex.append(len(subc.channels if is_noisy else subc.gates) + reindex[-1])
+
+        full_circuit = self._translate_circuit(circuit, translator_fn_name, qubit_order)
```
sergeisakov, line 883: The circuit gets translated twice, right? The circuit translation is very slow. Can the code be changed in such a way that the circuit is translated only once (if this is not too much work)?

sergeisakov (changes requested):
Good work!

pybind_interface/pybind_main.cpp
```diff
@@ -658,6 +723,37 @@ class SimulatorHelper {
...
+      subncircuit.num_qubits = ncircuit.num_qubits;
+      subncircuit.channels = std::vector<Channel<Gate>>(
+        ncircuit.channels.begin() + begin,
+        ncircuit.channels.begin() + end
+      );
+
+      Simulator simulator = factory.CreateSimulator();
+      StateSpace state_space = factory.CreateStateSpace();
+
+      result = NoisyRunner::RunOnce(params, subncircuit, seed, state_space,
```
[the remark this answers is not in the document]

qsimcirq/qsim_simulator.py
95-martin-orion: Memoization occurs in `_translate_circuit`, which this calls for each moment. This has the added benefit that repeated moments will use their memoized translations. (Note however that this does not affect `CircuitOperation`s with repetitions, as a given `CircuitOperation` exists inside a single moment)

qsimcirq/qsim_simulator.py
95-martin-orion: Single translation should be possible - I'll need to add circuit-composition code on the C++ side, but I don't think that should be too complicated. I'll include this in this PR.

95-martin-orion: Done: `compose[_noisy]_circuits` allows Python to merge two C++ circuits together.

95-martin-orion:
The tests pass locally for me, which suggests some version-specific misbehavior. I'll check it out.

sergeisakov:
> The tests pass locally for me, which suggests some version-specific misbehavior. I'll check it out.

The tests also pass locally for me.

qsimcirq/qsim_simulator.py
```diff
@@ -718,3 +857,33 @@ def _translate_circuit(
...
+        is_noisy: bool,
+    ):
+        """Translates a circuit by moments.
+
+        Returns:
+            A tuple containing the translated circuit and the gate indices
+            corresponding to moment boundaries.
+        """
+        reindex = [0]
+        full_circuit = qsim.NoisyCircuit() if is_noisy else qsim.Circuit()
```
sergeisakov, line 875: I would probably do this in a different way. I would modify `translate_cirq_to_qsim` and  `translate_cirq_to_qtrajectory` in qsim_circuit.py to return the gate indices corresponding to moment boundaries in addition to returning translated circuits. In this case, `qsim.compose_noisy_circuits` and `qsim.compose_circuits` are not required.

95-martin-orion: Sounds good to me, let's see how the CI tests feel about it.

95-martin-orion:
Okay, now I'm suspicious of issues external to this PR. The sum total of C++ changes between the latest commit and the last passing commit is a single line - and it's a comment. Meanwhile about half of the tests are now failing and at least one of them segfaults...

sergeisakov (approved):
LGTM

pybind_interface/pybind_main.cpp
```diff
@@ -559,6 +559,70 @@ class SimulatorHelper {
...
+        results[i] = helper.get_expectation_value(counts);
+        begin = end;
+      }
+      return results;
+    }
+
+    // Aggregate expectation values for noisy circuits.
+    for (unsigned i = 0; i < opsums_and_qubit_counts.size(); ++i) {
+      auto& counts = std::get<1>(opsums_and_qubit_counts[i]);
+      results[i] = std::vector<std::complex<double>>(counts.size(), 0);
```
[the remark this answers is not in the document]

PR #500, opened 2022-01-27
Remove temporary vector from trajectories.

sergeisakov:


lib/qtrajectory.h
```diff
@@ -279,50 +275,38 @@ class QuantumTrajectorySimulator {
...
 
         if (kop.unitary) continue;
 
-        // Apply the Kraus operator.
-        if (kop.ops.size() == 1) {
-          ApplyGate(simulator, kop.ops[0], state);
-        } else {
-          DeferOps(kop.ops, gates);
+        double prob = std::real(
+            simulator.ExpectationValue(kop.qubits, kop.kd_k.data(), state));
```
95-martin-orion, line 288: [No change needed] This is really clever! If I understand correctly, this works because:
```
prob(Ψ, K) = abs(K|Ψ>) ** 2 = <Ψ|K†K|Ψ> = EV(Ψ, K†K)
```
Is that correct?

95-martin-orion (approved):

lib/qtrajectory.h
sergeisakov: Yes, that is correct.

PR #501, opened 2022-02-04
Reuse quantum trajectory results.

sergeisakov:
Allows reusing quantum trajectory results to speed up simulations of circuits with weak noise and without measurements.

### DIFF
--- a/apps/qsim_qtrajectory_cuda.cu
+++ b/apps/qsim_qtrajectory_cuda.cu
@@ -251,6 +251,7 @@ int main(int argc, char* argv[]) {
   typename QTSimulator::Parameter param3;
   param3.max_fused_size = opt.max_fused_size;
   param3.verbosity = opt.verbosity;
+  param3.apply_last_deferred_ops = true;
 
   auto channel1 = AmplitudeDampingChannel<fp_type>(opt.amplitude_damp_const);
   auto channel2 = PhaseDampingChannel<fp_type>(opt.phase_damp_const);
@@ -262,6 +263,11 @@ int main(int argc, char* argv[]) {
   std::vector<std::vector<std::vector<std::complex<double>>>> results;
   results.reserve(opt.num_trajectories);
 
+  QTSimulator::Stat stat;
+
+  using CleanResults = std::vector<std::vector<std::complex<double>>>;
+  CleanResults primary_results(noisy_circuits.size());
+
   for (unsigned i = 0; i < opt.num_trajectories; ++i) {
     results.push_back({});
     results[i].reserve(noisy_circuits.size());
@@ -271,7 +277,6 @@ int main(int argc, char* argv[]) {
     auto seed = noisy_circuits.size() * (i + opt.traj0);
 
     for (unsigned s = 0; s < noisy_circuits.size(); ++s) {
-      std::vector<uint64_t> stat;
       if (!QTSimulator::RunOnce(param3, noisy_circuits[s], seed++,
                                 state_space, simulator, state, stat)) {
         return 1;
@@ -280,9 +285,22 @@ int main(int argc, char* argv[]) {
       results[i].push_back({});
       results[i][s].reserve(observables.size());
 
-      for (const auto& obs : observables) {
-        results[i][s].push_back(
-            ExpectationValue<IO, Fuser>(obs, simulator, state));
+      primary_results[s].reserve(observables.size());
+
+      if (stat.primary && !primary_results[s].empty()) {
+        for (std::size_t k = 0; k < observables.size(); ++k) {
+          results[i][s].push_back(primary_results[s][k]);
+        }
+      } else {
+        for (const auto& obs : observables) {
+          auto result = ExpectationValue<IO, Fuser>(obs, simulator, state);
+          results[i][s].push_back(result);
+
+          if (stat.primary) {
+            primary_results[s].push_back(result);
+            param3.apply_last_deferred_ops = false;
+          }
+        }
       }
     }
   }
--- a/lib/qtrajectory.h
+++ b/lib/qtrajectory.h
@@ -62,12 +62,14 @@ class QuantumTrajectorySimulator {
      * operators are sampled for each channel and there are no measurements
      * in the computational basis. This can be used to speed up simulations
      * of circuits with weak noise and without measurements by reusing
-     * the primary trajectory results. It is the client's responsibility
-     * to collect the primary trajectory results and to reuse them. There is
-     * an additional condition for RunBatch. In this case, the deferred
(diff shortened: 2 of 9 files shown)


lib/qtrajectory.h
```diff
@@ -55,6 +55,29 @@ class QuantumTrajectorySimulator {
...
     bool normalize_before_mea_gates = true;
+    /**
+     * If false, do not apply deferred operators after the main loop.
+     * This can be used to speed up simulations of circuits with weak noise
+     * and without measurements by reusing the results for the "primary" noise
+     * realization, that is the noise realization in which the  "primary"
+     * (highest probability) Kraus operators are sampled in each channel.
+     * The primary Kraus operators should be first in their respective channels.
+     */
+    bool apply_last_deferred_ops = true;
```
95-martin-orion, line 66: Does setting this to false reduce performance in the general case (i.e. non-weak noise)?

95-martin-orion (changes requested):
Haven't reviewed the tests yet - will handle that once I have a better understanding of how this optimization works.

lib/qtrajectory.h
```diff
@@ -55,6 +55,29 @@ class QuantumTrajectorySimulator {
      * If true, normalize the state vector before performing measurements.
      */
     bool normalize_before_mea_gates = true;
+    /**
+     * If false, do not apply deferred operators after the main loop.
```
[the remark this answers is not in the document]
sergeisakov: Added some clarifications. Defined "primary" Kraus operators in a different way.

lib/qtrajectory.h
```diff
@@ -310,11 +341,13 @@ class QuantumTrajectorySimulator {
       }
     }
 
-    if (!ApplyDeferredOps(param, num_qubits, simulator, gates, state)) {
-      return false;
-    }
+    if (apply_last_deferred_ops || !stat.primary) {
```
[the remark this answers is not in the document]
sergeisakov: Yes, if `apply_last_deferred_ops` is false and the primary trajectory is realized then this block is skipped and the state doesn't evolve through the deferred ops. This leads to a speedup. Note that `apply_last_deferred_ops` should be set to false after the first primary trajectory is realized (the state is fully evolved for the first primary trajectory and the required results are collected by the client).

lib/qtrajectory.h
```diff
@@ -368,10 +401,13 @@ class QuantumTrajectorySimulator {
...
 
-  static void CollectStat(bool collect_stat, uint64_t i,
-                          std::vector<uint64_t>& stat) {
+  static void CollectStat(bool collect_stat, uint64_t i, Stat& stat) {
     if (collect_stat) {
-      stat.push_back(i);
+      stat.samples.push_back(i);
+    }
+
+    if (i != 0) {
```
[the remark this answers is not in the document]
sergeisakov: Measurement gates in the computational basis (lines 262-284 in this file) are not compatible with `apply_last_deferred_ops=false`. Other measurements (represented by Kraus operators) are compatible with `apply_last_deferred_ops=false`.

lib/qtrajectory.h
sergeisakov: No, it doesn't reduce performance.

lib/qtrajectory.h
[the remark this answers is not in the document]
95-martin-orion: So during the first primary trajectory, results are collected for all subsequent realizations of that trajectory? I don't see where this is happening. For example, to me it looks like `RunBatch` measures the fully-evolved primary trajectory once, then measures the partially-evolved trajectory (i.e. without deferred ops) for all subsequent primary realizations, which doesn't seem like it would produce valid results.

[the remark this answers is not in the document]
sergeisakov: Yes, right after the first primary trajectory, results should be collected for all subsequent realizations of that trajectory. The user should do that, see tests and apps/qsim_qtrajectory_cuda.cu. `RunBatch` “measures” after every trajectory, i.e. it calls `measure` after every trajectory. The user should collect the required results in `measure` after the first primary trajectory (when `stat.primary` is set to true for the first time). We are not interested in the fully-evolved state for all subsequent primary trajectories because the state is always the same.

lib/qtrajectory.h
sergeisakov: This will produce invalid results in the general case because collecting and reusing the required results might not be implemented (as, say, in pybind_interface/pybind_main.cpp).

95-martin-orion: Got it, this makes sense now. I think it's really important to be clear that this is not a "flip the boolean for better performance" situation - let's separate out the line that calls this out:
```
/**
 * <existing comment, minus the line below>
 *
 * It is the client's responsibility to collect the primary trajectory results
 * and to reuse them.
 *\
```

lib/qtrajectory.h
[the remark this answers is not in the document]
95-martin-orion: > The user should collect the required results in `measure` after the first primary trajectory {...}

Ahh, this is the part I was missing (and now see that it's mentioned in the docstring). Thanks for bringing me up to speed.

apps/qsim_qtrajectory_cuda.cu
```diff
@@ -251,6 +251,7 @@ int main(int argc, char* argv[]) {
   typename QTSimulator::Parameter param3;
   param3.max_fused_size = opt.max_fused_size;
   param3.verbosity = opt.verbosity;
+  param3.apply_last_deferred_ops = true;
```
95-martin-orion, line 254: Should this be false? The other changes in this file seem built around the optimization in this PR.

95-martin-orion (changes requested):
Reviewed tests - all LGTM, with a couple small requests.

This is obviously applicable to deterministic `measure` methods like finding observables, but I think you could get sampling to work with it as well (e.g. by generating many samples when the primary is encountered, but only adding them to the output as more realizations appear).

lib/qtrajectory.h
sergeisakov: Done.

apps/qsim_qtrajectory_cuda.cu
sergeisakov: This should be true (this is the default value; just for self documenting). It is set to false below after the first primary trajectory.

sergeisakov:
> This is obviously applicable to deterministic `measure` methods like finding observables, but I think you could get sampling to work with it as well (e.g. by generating many samples when the primary is encountered, but only adding them to the output as more realizations appear).

I think the user should do that.

95-martin-orion (approved):

PR #504, opened 2022-02-09
Use qsim for repeated sampling

### LINKED ISSUE #503: Use qsim for repeated sampling
Highlighted in #502.

`_sample_measure_results` currently calls Cirq for repeated sampling of non-noisy circuits. This is a serviceable solution, but is somewhat less performant than simulating and sampling all on the qsim side. This is magnified by #502, where noiseless circuits on more than 32 qubits are forced to follow the repeated-simulation path.

We should replace the Cirq-dependent behavior with a method that handles this entirely in qsim.

95-martin-orion:
Fixes #503.

`*_sample` and `*_sample_final` need to be different functions because they have incompatible behaviors:
- `*_sample` (the existing methods) sample from measurements in a single execution of a circuit.
- `*_sample_final` (the new methods) sample repeatedly from the final state of a circuit.

### DIFF
--- a/pybind_interface/pybind_main.cpp
+++ b/pybind_interface/pybind_main.cpp
@@ -526,6 +526,15 @@ class SimulatorHelper {
     return helper.release_state_to_python();
   }
 
+  static std::vector<uint64_t> sample_final_state(
+      const py::dict &options, bool is_noisy, uint64_t num_samples) {
+    auto helper = SimulatorHelper(options, is_noisy);
+    if (!helper.is_valid || !helper.simulate(0)) {
+      return {};
+    }
+    return helper.sample(num_samples);
+  }
+
   template <typename StateType>
   static std::vector<std::complex<double>> simulate_expectation_values(
       const py::dict &options,
@@ -754,6 +763,11 @@ class SimulatorHelper {
     return result;
   }
 
+  std::vector<uint64_t> sample(uint64_t num_samples) {
+    StateSpace state_space = factory.CreateStateSpace();
+    return state_space.Sample(state, num_samples, seed);
+  }
+
   py::array_t<float> release_state_to_python() {
     StateSpace state_space = factory.CreateStateSpace();
     state_space.InternalToNormalOrder(state);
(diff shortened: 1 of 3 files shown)


pybind_interface/pybind_main.cpp
```diff
@@ -932,6 +946,16 @@ qtrajectory_simulate_moment_expectation_values(
 
 // Methods for sampling.
 
+std::vector<uint64_t> qsim_sample_final(
+    const py::dict &options, uint64_t num_samples) {
+  return SimulatorHelper::sample_final_state(options, false, num_samples);
+}
+
+std::vector<uint64_t> qtrajectory_sample_final(
```
sergeisakov, line 954: This function is not used, right?

sergeisakov (approved):
LGTM. Does it make sense to add some non-trivial test?

pybind_interface/pybind_main.cpp
95-martin-orion: Correct - repeated sampling from the final state of a noisy circuit isn't something we currently do. As it stands it doesn't seem like there's much harm in keeping it, but I'm happy to remove it if there's a concern about confusing users or if it's weighing down the library

sergeisakov: I think we can keep it.

sergeisakov:
> > LGTM. Does it make sense to add some non-trivial test?
> 
> I considered adding tests for #502, but a 33-qubit circuit of any depth would be excessive for a unit test. Other than that, the existing Python tests cover repeated sampling pretty thoroughly - is there anything extra you see that needs testing?

It seems that the tests are sort of trivial. For instance, the circuit in `test_qsim_run_vs_cirq_run` (line 405 in qsimcirq_test.py) just swaps the |0> amplitude and the `run` method always measures the same bit string. I was thinking of, say, a 4- or 5-qubit circuit with less trivial output and running a few thousands repetitions and comparing the measurement histogram with the actual probabilities. I am not sure if it makes sense to add such a test.

95-martin-orion:
The pybind call is simple enough that I think it's safe to rely on the C++ tests for verifying sampling behavior. The only new complexity on the Python side is in the bit-management of converting `raw_results` to `full_results` - I don't think a circuit with multiple possible sampling outcomes adds much coverage in this regard.

For reference, the existing `test_qsim_run_vs_cirq_run` uses a single-outcome circuit to specifically examine the behavior of that outcome. The result is asymmetric (0101) to check for bit-ordering errors, and multiple measurements are extracted to verify that mapping from raw data to keyed results is working.

sergeisakov:
SGTM

PR #509, opened 2022-02-22
Fix comparator bug which are not compliant to the strict weak ordering

hosang:
std::sort requries a strict weak ordering, but the comparator was reflexive (x < x).

### DIFF
--- a/lib/fuser_mqubit.h
+++ b/lib/fuser_mqubit.h
@@ -954,8 +954,11 @@ class MultiQubitGateFuser final : public Fuser<IO, Gate> {
 
                   if (ln != nullptr && rn != nullptr) {
                     return R()(ln->val->parent->time, rn->val->parent->time);
-                  } else {
+                  } else if (ln != rn) {
                     return ln != nullptr || rn == nullptr;
+                  } else {
+                    // Comparator needs to be irreflexive to get a strict weak ordering.
+                    return false;
                   }
                 });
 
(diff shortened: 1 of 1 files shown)


sergeisakov (changes requested):
Thank you for catching that. I think this can be simplified: the original line `return ln != nullptr || rn == nullptr;` can be just replaced with `return ln != nullptr;`.

hosang:
Made the code more concise, tested internally with patched in strict weak ordering test.

Needs to be `return ln != nullptr && rn == nullptr;` because it would become reflexive again for the case `ln != nulltr && rn != nulltr`.

lib/fuser_mqubit.h
```diff
@@ -955,7 +955,8 @@ class MultiQubitGateFuser final : public Fuser<IO, Gate> {
                   if (ln != nullptr && rn != nullptr) {
                     return R()(ln->val->parent->time, rn->val->parent->time);
                   } else {
-                    return ln != nullptr || rn == nullptr;
+                    // Comparator needs to be irreflexive to get a strict weak ordering.
+                    return ln != nullptr && rn == nullptr;
```
sergeisakov, line 959: I think `&& rn == nullptr` is redundant here because the case `ln != nullptr && rn != nullptr` is checked in line 995.

sergeisakov (changes requested):

lib/fuser_mqubit.h
hosang: Oooooh, finally got it, thanks, Sergei!

sergeisakov (approved):
LGTM

95-martin-orion:
Responsibility for mirroring this to the internal version is on me. Thanks @hosang!

hosang:
Thanks for the help, Orion and Sergei!

PR #519, opened 2022-03-16
Workaround Centos 8 EOL

### LINKED ISSUE #515: Update Docker images
Docker images for tutorials have fallen out of date, we need to update them to keep the tutorials working.

95-martin-orion:
Fixes #515.

Centos 8 has [reached end of life](https://www.centos.org/centos-linux-eol/), and our Docker builds are now breaking as a result. This PR uses the answer in [this StackOverflow question](https://stackoverflow.com/questions/70963985/error-failed-to-download-metadata-for-repo-appstream-cannot-prepare-internal) as a workaround to keep the builds functioning.

### DIFF
--- a/jupyter/Dockerfile
+++ b/jupyter/Dockerfile
@@ -3,6 +3,11 @@ FROM centos:8
 USER root
 # Install baseline
 
+RUN pushd /etc/yum.repos.d/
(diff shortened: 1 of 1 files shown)


95-martin-orion:
+Mike: This is the fix Doug mentioned that is necessary to use the qsim GCP scripts.

dstrain115 (approved):
Maybe add a comment to explain what these are for.  Otherwise, LGTM.

Issue #432, opened 2021-09-21
qsimcirq performance overhead


sergeisakov:
qsimcirq Python performance overhead is very high for small circuits. Say, for a 10-qubit circuit, it can be two orders of magnitude larger than the actual simulation time.

95-martin-orion:
[This profile graph](https://user-images.githubusercontent.com/40585662/162506819-306fdd30-45f2-4e1a-bd75-ea8e2dc0bc6f.png) is generated from a benchmark against 1000 circuits, each with 2 qubits and 4 gates. A notable culprit is the `getattr` method, which consumes ~10% of the runtime (compared to ~0.56% for the actual simulation)

PR #525, opened 2022-04-08
Reduce isinstance calls

95-martin-orion:
This PR improves Cirq-to-qsim translation performance by reducing time spent retrieving a Gate's type hierarchy.

The benchmark described in #432 produces [this profile graph](https://user-images.githubusercontent.com/40585662/162507198-85dd4b49-5777-444b-b1f3-ae85abd2bfe0.png) when run against this PR. In particular, note the significant reduction in `getattr` time (down to under 2% from 10+%).

### DIFF
--- a/qsimcirq/qsim_circuit.py
+++ b/qsimcirq/qsim_circuit.py
@@ -33,21 +33,30 @@ GATE_PARAMS = [
     "theta",
 ]
 
+
 def _translate_ControlledGate(gate: cirq.ControlledGate):
     return _cirq_gate_kind(gate.sub_gate)
+
+
 def _translate_IdentityGate(gate: cirq.IdentityGate):
     # Identity gates will decompose to no-ops.
     pass
+
+
 def _translate_XPowGate(gate: cirq.XPowGate):
     # cirq.rx also uses this path.
     if gate.exponent == 1 and gate.global_shift == 0:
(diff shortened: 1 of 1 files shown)


qsimcirq/qsim_circuit.py
```diff
@@ -34,103 +34,182 @@
...
+    return qsim.kMeasurement
+
+
+TYPE_TRANSLATOR = {
+    cirq.ControlledGate: _translate_ControlledGate,
+    cirq.XPowGate: _translate_XPowGate,
+    cirq.YPowGate: _translate_YPowGate,
+    cirq.ZPowGate: _translate_ZPowGate,
+    cirq.HPowGate: _translate_HPowGate,
+    cirq.CXPowGate: _translate_CXPowGate,
```
sergeisakov, line 177: It seems `CZPowGate` is missing here.

sergeisakov (changes requested):

qsimcirq/qsim_circuit.py
95-martin-orion: Added (likely a copy-paste error) with negligible effect on benchmark results.

sergeisakov (approved):
LGTM. The Kokoro build failed on Windows. It's probably not related to this PR.

Issue #495, opened 2022-01-04
Can't import qsimcirq on MacOS ARM64


honamnguyen:
`pip install qsimcirq` works on my Mac with the M1 chip but then `import qsimcirq` gives the following error
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/honamnguyen/miniconda3/envs/qcircuit/lib/python3.9/site-packages/qsimcirq/__init__.py", line 17, in <module>
    from qsimcirq import qsim_decide
ImportError: dlopen(/Users/honamnguyen/miniconda3/envs/qcircuit/lib/python3.9/site-packages/qsimcirq/qsim_decide.cpython-39-darwin.so, 0x0002): tried: '/Users/honamnguyen/miniconda3/envs/qcircuit/lib/python3.9/site-packages/qsimcirq/qsim_decide.cpython-39-darwin.so' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e')), '/usr/local/lib/qsim_decide.cpython-39-darwin.so' (no such file), '/usr/lib/qsim_decide.cpython-39-darwin.so' (no such file)
```
I found these threads [here](https://github.com/nmwsharp/robust-laplacians-py/issues/7) and [here](https://github.com/nmwsharp/robust-laplacians-py/pull/8) that seem to have a solution which is way over my head but might be useful. If not, I'd also appreciate any suggestion for a work-around. Thanks!

Akimasa11:
I have the same problem where I could install the package but encounter the below error when importing it. 
`ImportError: dlopen(/Users/akimasaihara/opt/anaconda3/lib/python3.8/site-packages/qsimcirq/qsim_decide.cpython-38-darwin.so, 0x0002): Library not loaded: /usr/local/opt/libomp/lib/libomp.dylib
  Referenced from: /Users/akimasaihara/opt/anaconda3/lib/python3.8/site-packages/qsimcirq/qsim_decide.cpython-38-darwin.so
  Reason: tried: '/usr/local/opt/libomp/lib/libomp.dylib' (no such file), '/usr/local/lib/libomp.dylib' (no such file), '/usr/lib/libomp.dylib' (no such file), '/opt/homebrew/Cellar/libomp/13.0.1/libomp.dylib' (no such file), '/usr/local/lib/libomp.dylib' (no such file), '/usr/lib/libomp.dylib' (no such file)`

When I tried using a virtual environment I got the following error during installation.
[...]

PR #530, opened 2022-05-13
DisARM qsim

95-martin-orion:
As noted in #495, qsim wheels are not compatible with MacOS M1. This PR removes the `universal2` wheels from the test and release workflows so that M1s do not install a known-invalid wheel.

Some complications will persist: M1s could still install wheel published for older qsim versions, and building qsim from source is also broken on M1 devices. However, this at least stops us from continuing to advertise a functioning ARM-compatible qsim wheel in subsequent releases.

### DIFF
--- a/.github/workflows/release_wheels.yml
+++ b/.github/workflows/release_wheels.yml
@@ -15,12 +15,7 @@ jobs:
           - os: macos-10.15
             name: mac
             cibw:
-              build: "cp36* cp37* cp38*"
-          - os: macos-10.15
-            name: mac-arm
-            cibw:
-              arch: universal2
-              build: "cp39*"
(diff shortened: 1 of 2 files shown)


mpharrigan (approved):
bonus points for pun.

Is there a way to publish like a tombstone package or something that will prevent pip from happily installing the already-published wheels?

95-martin-orion:
> bonus points for pun.
> 
> Is there a way to publish like a tombstone package or something that will prevent pip from happily installing the already-published wheels?

It seems likely that there's something for this, but singling out a specific wheel for a specific architecture requesting it might get finicky. I'll add a note to the issue to see if we can flag this.

sergeisakov (approved):
LGTM

PR #531, opened 2022-05-27
Add support for repeated keys in QSimSimulator

maffoo:


qsimcirq/qsim_simulator.py
```diff
@@ -323,30 +323,35 @@ def _sample_measure_results(
...
+            if key in num_qubits_by_key:
+                if n != num_qubits_by_key[key]:
+                    raise ValueError(
+                        f"repeated key {key!r} with different numbers of qubits: "
+                        f"{num_qubits_by_key[key]} != {n}"
+                    )
+            else:
+                num_qubits_by_key[key] = n
+            if key not in bounds:
+                bounds[key] = (current_index, current_index + n)
```
95-martin-orion, line 345: `bounds` is used at the bottom of this function to determine how to map measurement results from qsim's raw result output to Cirq's keyed results dictionary. In order for this to work, we'll need bounds stored for each instance of the given measurement key.

qsimcirq_tests/qsimcirq_test.py
```diff
@@ -67,7 +67,7 @@ def test_repeated_keys():
         cirq.measure(q, key="m"),
     )
     result = qsimcirq.QSimSimulator().run(circuit, repetitions=10)
-    assert result.records["m"] == (10, 3, 1)
+    assert result.records["m"].shape == (10, 3, 1)
```
95-martin-orion, line 70: Could you add tests for reusing measurement keys with:
- different qubits, different moment (ok)
- different qubits, same moment (error)
- different # of qubits (error)
- CircuitOperation(use_repetition_ids=False) (ok)

95-martin-orion (changes requested):
Can this be marked ready for review (non-draft)?

PR #536, opened 2022-07-06
Sanity check

95-martin-orion:
An effort to verify that the macOS build is _actually_ broken before merging #419 to fix it.


95-martin-orion:
Confirmed: qsim is actually broken at master and requires a fix.

PR #538, opened 2022-07-08
Add python 3.10 to wheel list

### LINKED ISSUE #537: Can't install later than qsim 0.9.5 with python 3.10

95-martin-orion:
Fixes #537.

There's a pretty good chance that this breaks something else, but testing is the easiest way to find out what :)


95-martin-orion:
N.B. that a new version (0.13.1?) will be required to actually release a py3.10 wheel.

dstrain115 (approved):

PR #541, opened 2022-07-08
bump cibuildwheel version

95-martin-orion:
Wow this was old.


95-martin-orion:
@MichaelBroughton turns out the python3.10 wheels won't build unless cibuildwheel is updated.

MichaelBroughton (approved):

PR #544, opened 2022-07-11
Update version to 0.13.2

### LINKED ISSUE #537: Can't install later than qsim 0.9.5 with python 3.10

95-martin-orion:
Picking up #541 to get a Python 3.10 wheel out.

Fixes #537.


MichaelBroughton (approved):

95-martin-orion:
...checks still don't show a 3.10 wheel being built, and I'm at a loss as to why. Will continue to investigate.

95-martin-orion:
Solved in #549.

PR #546, opened 2022-07-12
Upgrade to latest setup-python action

95-martin-orion:
Yet another attempt at #537. The older version of `setup-python` may not have installed python 3.10 by default.


95-martin-orion:
...still no python 3.10 wheel.

95-martin-orion:
Dropping in favor of the investigation in #547.

PR #547, opened 2022-07-12
sanity check

95-martin-orion:
DO NOT MERGE

Checking if CIBW changes are actually being processed.

### DIFF
--- a/.github/workflows/release_wheels.yml
+++ b/.github/workflows/release_wheels.yml
@@ -29,6 +29,7 @@ jobs:
               build: "cp36-win_amd64 cp37-win_amd64 cp38-win_amd64 cp39-win_amd64 cp310-win_amd64"
     env:
       CIBW_BUILD: "${{ matrix.cibw.build || '*' }}"
+      CIBW_SKIP: "*musllinux*"
       CIBW_ARCHS: "${{ matrix.cibw.arch || 'auto' }}"
       CIBW_MANYLINUX_X86_64_IMAGE: "${{ matrix.cibw.manylinux_image }}"
       CIBW_BEFORE_BUILD_MACOS: "brew install libomp llvm"
--- a/.github/workflows/testing_wheels.yml
+++ b/.github/workflows/testing_wheels.yml
@@ -20,18 +20,18 @@ jobs:
           - os: macos-10.15
             name: mac
             cibw:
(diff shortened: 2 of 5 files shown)


95-martin-orion:
This was enlightening: apparently `cp310*` does not match any build configs. This disagrees with [cibw docs](https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip), which suggests that somehow we don't have access to the latest config.

...and looking back at [cibw 2.0.1 docs](https://github.com/pypa/cibuildwheel/blob/v2.0.1/docs/options.md), it's clear that this is the issue. 2.2.2 seems to be the earliest version with explicit 3.10 configs.

95-martin-orion:
Good news: we've added Python 3.10!

Bad news: everything breaks with Python 3.10!

95-martin-orion:
[Thsi SO issue](https://stackoverflow.com/questions/69954587/no-blas-lapack-libraries-found-when-installing-scipy-on-macos/72034993) seems relevant

95-martin-orion:
Trying `numpy==1.16.5` based on [scipy-numpy version matching](https://docs.scipy.org/doc/scipy/dev/toolchain.html#numpy). The scipy version seems to be our complicating factor, but we don't have much control over it from qsim.

95-martin-orion:
Trying an even newer numpy version since [numpy.org](https://numpy.org/news/) indicates numpy did not support Python 3.10 until sometime in the numpy 1.21 era. (Might need to bump to 1.22 if this still fails)

95-martin-orion:
Ubuntu failure was only on musllinux, which we haven't previously used. Excluding it should resolve ubuntu.

Windows failure log was inscrutable - rerunning to test for transience.

95-martin-orion:
Windows issue persists, but matches a similar issue described [here](https://github.com/pybind/pybind11/issues/1342). Applying the fix described there.

Also disabling the musllinux build with `CIBW_SKIP`.

95-martin-orion:
Ubuntu error is resolved, but Windows continues to fail after adding `-DMS_WIN64`. Unclear if the flag was actually used in compilation; will continue to push on this solution.

95-martin-orion:
Okay, it appears this is related to the build somehow selecting the ancient `pybind11==2.2.4` (from _Sept. 2018_). Trying to force a more recent pybind version to resolve things.

PR #549, opened 2022-07-13
Advance versions for compatibility

### LINKED ISSUE #537: Can't install later than qsim 0.9.5 with python 3.10

95-martin-orion:
Fixes #537.

Bumps forward several pinned versions to get qsim building with Python 3.10.

### DIFF
--- a/.github/workflows/release_wheels.yml
+++ b/.github/workflows/release_wheels.yml
@@ -15,18 +15,18 @@ jobs:
           - os: macos-10.15
             name: mac
             cibw:
-              build: "cp36* cp37* cp38* cp39* cp310*"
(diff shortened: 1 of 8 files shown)


MichaelBroughton (approved):

rht:
Quick question: will the Docker image in gcr.io be updated with 3.7 soon (within a day) after this PR gets merged?

95-martin-orion:
> Quick question: will the Docker image in gcr.io be updated with 3.7 soon (within a day) after this PR gets merged?

Our process for updating the Docker images is unfortunately quite a bit more manual than the pypi release - if this is a priority for you, please open an issue and I'll get to it as soon as I can.

PR #552, opened 2022-07-15
Convert to Debian

95-martin-orion:
Switches the base Docker image from CentOS 8 (which is past end-of-life) to Debian.

This is necessary as CentOS 8 does not support Python > 3.6, but both qsim and Cirq now require Python 3.7+. Tested in Google Cloud Shell and verified that the new image gets the latest qsim+cirq versions (0.13.3 and 0.15.0, respectively).

### DIFF
--- a/jupyter/Dockerfile
+++ b/jupyter/Dockerfile
@@ -1,33 +1,24 @@
 # Base OS
-FROM centos:8
+FROM debian:bullseye
 USER root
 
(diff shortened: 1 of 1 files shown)


95-martin-orion:
CC @jrossthomson

dstrain115 (approved):
Assuming you have tested the docker image and it actually works, this sounds fine to me.

PR #554, opened 2022-07-18
Prevent duplicate noise addition

### LINKED ISSUE #553: A strange problem about the QSimSimulator
Hi there,

I encountered a strange problem when using qsim. Below is the script that can reproduce the error. Basically, I first generate the set of all single-qubit Clifford gates, which is `_single_qubit_clifford`. Then, I define a circuit by adding gates layer by layer; one layer of single-qubit Clifford gate followed by a layer of CNOTs and so on. Besides, I also added errors after each layer. This will be perfectly fine to the qsim simulator. 

But for some reason, I need to insert some addition gates in the middle of the circuit, which is `cirq.X(qubits[0])` in the example. Then, it return an error message, saying `ValueError: could not broadcast input array from shape (0,) into shape (4,)`. Below is the case that I found will be no problem:
- If the extra gate is removed, then that's OK;
- If all errors are removed, then that's OK;
- If the Pauli-X gate is merged with the gate before it, using the `cirq.MatrixGate` for construction, then that's also fine.

I tried to look into this problem for several hours, but I couldn't figure out what is going on exactly, only to discover how to work around it by the 'merge' method. Any idea what's happening? Thanks! 

```python
import cirq, qsimcirq
import itertools

# generate all single-qubit clifford gates
_pauli_list = (cirq.I, cirq.X, cirq.Y, cirq.Z)
_bools = (False, True)

def _all_rotation_pairs():
    for px, flip_x, pz, flip_z in itertools.product(_pauli_list[1:], _bools, _pauli_list[1:], _bools):
        if px == pz:
            continue
        yield (px, flip_x), (pz, flip_z)

def _all_clifford_gates():
    for trans_x, trans_z in _all_rotation_pairs():
        yield cirq.SingleQubitCliffordGate.from_xz_map(trans_x, trans_z)

_single_qubit_clifford = tuple(_all_clifford_gates())

# simulators
dm_sim = cirq.DensityMatrixSimulator()
qsim_sim = qsimcirq.QSimSimulator()

# define the circuit
qubits = cirq.LineQubit.range(4)
error = cirq.phase_flip(0.02)
circ = cirq.Circuit(
    _single_qubit_clifford[0].on_each(*qubits),
    cirq.X(qubits[0]),
    error.on_each(*qubits),
    cirq.CNOT(qubits[0], qubits[1]),
    cirq.CNOT(qubits[2], qubits[3]),
    error.on_each(*qubits),
    _single_qubit_clifford[4].on_each(*qubits),
    error.on_each(*qubits), 
    cirq.measure_each(*qubits)
)

# run
print(dm_sim.run(circ))
print(qsim_sim.run(circ))
```

95-martin-orion:
Fixes #553.

Prior to this change, if a moment contained both (1) a noiseless gate that decomposes into multiple gates, and (2) a noisy gate, the circuit translator would (incorrectly) add a copy of the noisy gate for every gate produced by the noiseless gate.

Fortunately, because this generates a pair of gates in the translated circuit which overlap in time and targeted qubits, there is no risk of this having polluted previous experiments - anyone hitting this issue would see an error rather than incorrect results.

### DIFF
--- a/qsimcirq/qsim_circuit.py
+++ b/qsimcirq/qsim_circuit.py
@@ -473,6 +473,9 @@ class QSimCircuit(cirq.Circuit):
                     time = time_offset + gi
                     add_op_to_circuit(qsim_op, time, qubit_to_index_dict, qsim_ncircuit)
                     gate_count += 1
+                # Only gates decompose to multiple time steps.
+                if gi > 0:
+                    continue
                 # Handle mixture output.
                 for mixture in ops_by_mix:
                     mixdata = []


95-martin-orion:
TODO: add test to verify this fixes the behavior (verified manually)

Issue #543, opened 2022-07-11
Upgrade for compatibility with Cirq 1.0


95-martin-orion:
Several deprecated parts of Cirq are being removed in preparation for its 1.0 release. In order for qsim to remain compatible, we need to upgrade its behavior to match.

The nightly compatibility check is a useful indicator for what's currently broken.

95-martin-orion:
Sample nightly compatibility run: [link](https://github.com/quantumlib/qsim/runs/7312352387?check_suite_focus=true)

maffoo:
Fixed by #555.

PR #555, opened 2022-07-18
Update for compatibility with cirq 1.0

maffoo:
WIP for #543. There's some messiness here around dealing with control values for controlled gates, which are still in flux in cirq so this will have to be updated when those changes land (https://github.com/quantumlib/Cirq/pull/5788).

### DIFF
--- a/qsimcirq/__init__.py
+++ b/qsimcirq/__init__.py
@@ -53,12 +53,7 @@ qsim_gpu = _load_qsim_gpu()
 qsim_custatevec = _load_qsim_custatevec()
 
 from .qsim_circuit import add_op_to_opstring, add_op_to_circuit, QSimCircuit
-from .qsim_simulator import (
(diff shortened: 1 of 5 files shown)


95-martin-orion:
Locally verified that this is compatible with the latest Cirq prerelease (cirq-0.16.0.dev20220718175125).

Thanks @maffoo for this patch fix! Because of how the CI tests are set up, we'll need to wait for the Cirq release to merge this quietly. If this becomes urgent, we can do a break-glass merge before the release and realign with 1.0 afterwards.

maffoo:
@95-martin-orion thanks for trying this out. Not particularly urgent from my perspective, so I'm happy to wait until the cirq 1.0 release is out.

95-martin-orion (approved):
LGTM, may the CI find us worthy

PR #562, opened 2022-08-26
Defer to HTC Toolkit for terraforming

95-martin-orion:
Our multinode tutorial is now based on HTC Toolkit, so these are no longer needed.

Deleting them to avoid confusion.


95-martin-orion:
CC @jrossthomson @tpdownes

tpdownes:
@95-martin-orion removing this code would provide a little more clarity for customers/users. Any progress?

sergeisakov (approved):

Issue #526, opened 2022-04-11
Windows build broken


95-martin-orion:
The Windows builds on Kokoro are breaking with a number of disconcerting messages:
```
> bazel test --test_output=errors --copt=/D_USE_MATH_DEFINES tests:all
ERROR: C:/tools/msys64/home/kbuilder/_bazel_kbuilder/yic25m7j/external/local_config_cc/BUILD:274:24:
unterminated string literal at eol
{...}
Target '//lib:vectorspace' depends on toolchain '@local_config_cc//:cc-compiler-x64_windows',
which cannot be found
```
These do not appear in previous passing runs. This failure is blocking submit (#521).

95-martin-orion:
This appears to be the result of changes in the Tensorflow toolchain location - hoping to have a fix out later today.

95-martin-orion:
In the absence of a quick solution for this, we have opted to disable the qsim Windows build to unblock other PRs. We will continue working with the TFQ team to resolve the issue, but for the time being Windows users are advised to install the most recent release prior to the issue (`qsimcirq==0.12.1`).

PR #563, opened 2022-09-23
Mark AVX- and SSE-using targets as incompatible with non-x86 platforms.

lexi-nadia:


95-martin-orion (approved):
Changes LGTM. I've triggered the Kokoro build to test the C++ build across Linux/Mac/Windows - if there are any odd edge cases that should surface them for us.

95-martin-orion:
@lexi-nadia the Kokoro logs should be visible from the details link - let me know if you don't have access.

From those logs:
```
//lib:statespace_avx: no such attribute 'target_compatible_with' in 'cc_library' rule
{ ...similar logs for other rules... }
```

95-martin-orion:
Summary of offline discussion: `target_compatible_with` was not added until bazel v4.0. While the qsim bazel version could plausibly advance (although see #526 for issues around that), it cannot advance past the bazel version used in Tensorflow Quantum, which is 3.7.2.

Instead, these changes will be applied to the pipeline used to construct the Google-internal mirror of qsim.

PR #564, opened 2022-09-23
Exclude AVX- and SSE-using source files from non-x86 builds

lexi-nadia:


lexi-nadia:
This is a second attempt. This PR does not use `target_compatible_with`; instead, it only uses `select`, which is supported by the version of Bazel that qsim currently uses.

95-martin-orion:
```
ERROR: Analysis of target '//apps:qsim_amplitudes' failed; build aborted:
       no such package '@platforms//cpu': The repository '@platforms' could not be resolved
```

PR #566, opened 2022-09-27
Catch identities in expectation value calcs

### LINKED ISSUE #565: QSimSimulator.simulate_expectation_values for identity operators is wrong
QSimSimulator.simulate_expectation_values returns wrong results for identity operators in observables.
As the following example shows, identity operators in observables seem to be regarded as zero matrices.

This is a straightforward example.
```
import cirq
import qsimcirq

circuit = cirq.Circuit()
q = cirq.LineQubit.range(1)
ps = cirq.PauliString(cirq.I(q[0]))

sim = qsimcirq.QSimSimulator()
evs = sim.simulate_expectation_values(
    circuit, ps)
print("qsim", evs)

cirq_sim = cirq.Simulator()
evs = cirq_sim.simulate_expectation_values(
    circuit, ps)
print("cirq", evs)
```

And the result is here.
```
qsim [0j]
cirq [(1+0j)]
```

95-martin-orion:
Fixes #565.

`cirq.PauliString` does not include identities in its `.items()` representation. Changing this is difficult due to how Cirq handles Paulis; instead, it is simpler to treat empty `PauliString`s as identities in qsim.

### DIFF
--- a/pybind_interface/pybind_main.cpp
+++ b/pybind_interface/pybind_main.cpp
@@ -797,6 +797,9 @@ class SimulatorHelper {
       const auto& opsum_qubits = std::get<1>(opsum_qubit_count_pair);
       if (opsum_qubits <= 6) {
         results.push_back(ExpectationValue<IO, Fuser>(opsum, simulator, state));
+      } else if (opsum_qubits == 0) {
+        // This represents an identity, whose EV is always 1.
+        results.push_back(1);
       } else {
         Fuser::Parameter params;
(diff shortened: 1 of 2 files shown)


95-martin-orion:
The reason we haven't seen this before is that identities were not supported in `cirq.PauliString` until recently: https://github.com/quantumlib/Cirq/pull/5621

sergeisakov (approved):
LGTM

PR #579, opened 2023-01-12
Add global phase gate.

sergeisakov:


95-martin-orion:
Before I dig into review: this PR impacts more of qsim than I would have expected. Is this primarily due to adding support for zero-qubit gates, or were there other complications that expanded the change?

sergeisakov:
Yes, this is due to adding support for zero-qubit gates. Perhaps the only non-trivial change in this PR is addition to the fusers.

lib/fuser_basic.h
```diff
@@ -212,6 +220,8 @@ class BasicGateFuser final : public Fuser<IO, Gate> {
           gates_lat[gate.qubits[0]].push_back(&gate);
           gates_lat[gate.qubits[1]].push_back(&gate);
           gates_seq.push_back(&gate);
+        } else {
```
95-martin-orion, line 223: Is there any case where this `else` triggers for gates with >2 qubits?

lib/simulator_custatevec.h
```diff
@@ -80,23 +98,30 @@ class SimulatorCuStateVec final {
...
                            const fp_type* matrix, State& state) const {
-    std::vector<int32_t> control_bits;
-    control_bits.reserve(cqs.size());
-
-    for (std::size_t i = 0; i < cqs.size(); ++i) {
-      control_bits.push_back((cmask >> i) & 1);
-    }
+    if (qs.size() == 0) {
+      IO::errorf(
+          "error: controlled global phase gate is not implemented %s %d\n",
```
95-martin-orion, line 103: Is this intended for this PR, planned for a future PR, or not on the roadmap?

95-martin-orion (approved):

lib/fuser_basic.h
sergeisakov: I think there is no such a case. The `else` in line 206 triggers for gates with >2 qubits

lib/simulator_custatevec.h
sergeisakov: It is not on the roadmap.

rht:
Thank you so much for the PR!

PR #580, opened 2023-01-13
use_gpu in custatevec options

95-martin-orion:
From offline discussion:
> GPU acceleration is not really enabled with just gpu_mode=1.


sergeisakov (changes requested):
Lines 1389, 1412 and 1431 should also be modified.

95-martin-orion:
> Lines 1389, 1412 and 1431 should also be modified.

Done.

sergeisakov (approved):

PR #581, opened 2023-01-18
Update version to v0.15.0

95-martin-orion:
This minor version update is for the new global phase gate feature, added in #579.

Additionally, there have been a number of small patch fixes since 0.14.1:
- Fix array overflow for small input vectors (#572 and #582)
- Fix weighted-identity expectation values (#577 and #578)
- Fix custatevec not being enabled on request (#580)


95-martin-orion:
On hold while GPU analogue of #572 is rolled out.

sergeisakov (approved):

PR #582, opened 2023-01-19
Fix array overflow for small input vectors.

sergeisakov:


95-martin-orion (approved):
This is a GPU follow-up to #572.

PR #588, opened 2023-02-22
Fix expectation values.

sergeisakov:


95-martin-orion (changes requested):
Could you add a test to capture the issue this is fixing?

sergeisakov:
> Could you add a test to capture the issue this is fixing?

Done.

PR #595, opened 2023-03-17
Upgrade to TFv2.11 and add CUDA support

jaeyoo:


jaeyoo:
https://github.com/quantumlib/qsim/pull/596 resolves this together. closed.

PR #606, opened 2023-04-29
Fix simmux_gpu

QuantumJaeYoo:


lib/simmux_gpu.h
```diff
@@ -18,12 +18,12 @@
 #ifdef __CUSTATEVEC__
 # include "simulator_custatevec.h"
   namespace qsim {
-    using SimulatorGpu = SimulatorCuStateVec;
+    using SimulatorGpu = SimulatorCuStateVec<>;
```
95-martin-orion, line 21: I assume this is necessary to use the default template argument? (i.e. `FP=float`)

95-martin-orion (approved):
One note to confirm, otherwise LGTM.

lib/simmux_gpu.h
jaeyoo: In fact, that is just renaming of the template. I forgot `<>`, so it says compile error.

Because `simmux.h` itself didn't provide default argument of data type FP, and only provides `For` for partial argument passing, I just followed the same template renaming.

So, users can use `SimulatorGpu<float>`. Thank you for checking this!

But due to CLA issues, let me merge two PRs into one by corp account.

PR #609, opened 2023-05-03
Update `_version.py` to `0.16.2` in anticipation for new qsim release.

vtomole:

PR #610, opened 2023-05-03
Update `cibuildwheel` version

vtomole:
See https://github.com/quantumlib/qsim/pull/609#issuecomment-1533660555

### DIFF
--- a/.github/workflows/release_wheels.yml
+++ b/.github/workflows/release_wheels.yml
@@ -47,7 +47,7 @@ jobs:
       - uses: actions/setup-python@v2
 
       - name: Install cibuildwheel and twine
-        run: python -m pip install cibuildwheel==2.2.2
+        run: python -m pip install cibuildwheel==2.12.3
 
       - name: Install requirements
         run: python -m pip install -r requirements.txt
--- a/.github/workflows/testing_wheels.yml
+++ b/.github/workflows/testing_wheels.yml
@@ -52,7 +52,7 @@ jobs:
       - uses: actions/setup-python@v2
 
       - name: Install cibuildwheel and twine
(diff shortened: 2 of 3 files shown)


95-martin-orion:
To remind myself: we should be able to view the generated wheels in the `Build all wheels for testing` logs to confirm that this does what we need it to do.

95-martin-orion:
From the error logs, it looks like we also need to update our pybind version. `pybind11==2.10.1` is the first version with python 3.11 support ([docs](https://pybind11.readthedocs.io/en/stable/changelog.html#version-2-10-1-oct-31-2022)), and we set the version [here](https://github.com/quantumlib/qsim/blob/30695ff8a26e817594822bd5beccbfe8a62e41dd/pybind_interface/GetPybind11.cmake#L3).

vtomole:
Updated to minimum pybind version to `2.10.1`.

vtomole:
@95-martin-orion Looks like this PR generates a `qsimcirq-0.16.2-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`. Will we need to cut a new release once this is in?

95-martin-orion:
> @95-martin-orion Looks like this PR generates a `qsimcirq-0.16.2-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`. Will we need to cut a new release once this is in?

Yup, same as before. All PRs generate wheels - and in theory you could use them, if sufficiently motivated - but they aren't included in `pip install qsimcirq` until they're built into a release.

95-martin-orion (approved):

vtomole:
@95-martin-orion No, thank you :smile_cat:

PR #616, opened 2023-05-31
Update to latest MacOS runner version

### LINKED ISSUE #615: Update MacOS builds
The GHA list of [supported versions](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) no longer includes MacOS 10.15, which is causing our builds to hang waiting for runners that no longer exist. We need to update to at least `macos-11` to resolve this.

95-martin-orion:
Fixes #615.

The root cause of this issue was identified from https://stackoverflow.com/q/70959954

### DIFF
--- a/.github/workflows/release_wheels.yml
+++ b/.github/workflows/release_wheels.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: macos-10.15
+          - os: macos-12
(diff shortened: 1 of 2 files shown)


95-martin-orion:
CC @eliottrosenberg

sergeisakov (approved):

Issue #625, opened 2023-09-22
`Bazel / Sanitizer tests (msan)` CI failes due to C++ test failure


NoureldinYosri:
It's not clear why this is happening since other Bazel/Sanitizer tests (basic, sse, avx) pass. The test failes even on PR that don't touch the C++ code.

Example: https://github.com/quantumlib/qsim/actions/runs/5914985907/job/16076067845?pr=621

95-martin-orion:
Unlike Cirq, qsim runs its entire CI suite on all PRs. This is mostly a function of qsim having far fewer active contributors than Cirq, rather than any technical limitation; if someone is interested in segmenting qsim CI, they are welcome to take it on (though bear in mind that it is not as friendly as Cirq CI).

95-martin-orion:
As for why msan fails and the others don't: msan is a "memory sanitizer test". It's not testing the tests themselves - although it will fail if the tests fail - it's testing whether the memory management in the tests is sound.

NoureldinYosri:
@95-martin-orion  so this CI fails because there is an open PR that modifies the C++ code?

95-martin-orion:
> @95-martin-orion so this CI fails because there is an open PR that modifies the C++ code?

No - the CI for each PR is self-contained, and should not be affected by other open, unmerged PRs. Given that the PR in question doesn't affect the test, the issue is most likely due to unpinned dependencies advancing to a version which qsim is not compatible with, or which has internal memory leaks. "Dependencies" here could include python dependencies, the C++ compiler, or the sanitizers themselves.

Searching for the error (`lsan_init_is_running`) finds https://github.com/google/sanitizers/issues/682, which suggests that it may be an LLVM bug.

95-martin-orion:
qsim is no longer my main project, so I don't have the availability to root-cause this. @sergeisakov Do you have bandwidth to investigate this? Otherwise, we would happily accept outside contributions to fix the issue.

sergeisakov:
I don't really know how to fix this issue. It's not related to the C++ code.

NoureldinYosri:
The test breaks at compilation step due to an update to the LLVM library https://bugs.llvm.org/show_bug.cgi?id=27310. fix is to update libc-bin which I added to my PR #623

Issue #624, opened 2023-09-22
`Docker / Test on ubuntu-latest` CI failes due to numpy conflict


NoureldinYosri:
by the time https://github.com/quantumlib/qsim/blob/235ae2fc039fb4a98beb4a6114d10c7f8d2070f7/pybind_interface/Dockerfile#L9 gets executed a numpy version is installed by pybind11 at https://github.com/quantumlib/qsim/blob/235ae2fc039fb4a98beb4a6114d10c7f8d2070f7/pybind_interface/Dockerfile#L5

The `--force` doesn't help because this is a conflict between a pip backage (for cirq) and a system package.

Example https://github.com/quantumlib/qsim/actions/runs/5914985909/job/16076067832?pr=621

PR #628, opened 2023-10-05
fix Docker and Bazel CIs

NoureldinYosri:
fixes https://github.com/quantumlib/qsim/issues/624
the docker CI failes due to a mismatch between system packages scientific libraries (e.g. numpy) installed by `python3-pybind11` and those required by Cirq. Fix by allowing Cirq to override those packages' versions. This is the same as we had before with the `--force` flag which doesn't seem to work anymore, instead I'm using `a virtualenv

---

fixes https://github.com/quantumlib/qsim/issues/625
these were broken due to an update to the LLVM library which broke the compilation step https://bugs.llvm.org/show_bug.cgi?id=27310. Fix is to upgrade libc-bin.

### DIFF
--- a/.github/workflows/bazeltest.yml
+++ b/.github/workflows/bazeltest.yml
@@ -60,6 +60,10 @@ jobs:
       - name: Install requirements
         run: |
           python3 -m pip install -r requirements.txt
+      - name: Upgrade libc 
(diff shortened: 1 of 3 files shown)


95-martin-orion (approved):
All checks are passing. Thank your @NoureldinYosri for stepping in to resolve these issues!

PR #630, opened 2023-10-20
Adding Qsim confusion matrix runtime exception

angelo-laskaris:
QSim currently does not support confusion matrices on measurement gates. See https://github.com/quantumlib/Cirq/issues/6305 for more info. While work is done to support confusion matrices surfacing a runtime exception.

### DIFF
--- a/qsimcirq/qsim_simulator.py
+++ b/qsimcirq/qsim_simulator.py
@@ -213,7 +213,7 @@ class QSimSimulator(
         """Run a simulation, mimicking quantum hardware.
 
         Args:
-            program: The circuit to simulate.
+            circuit: The circuit to simulate.
             param_resolver: Parameters to run with the program.
             repetitions: Number of times to repeat the run.
 
--- a/docs/choose_hw.md
+++ b/docs/choose_hw.md
@@ -258,7 +258,7 @@ are “embarrassingly parallelizable”, there is an automated workflow for
 distributing these trajectories over multiple nodes. A simulation of many
 noiseless circuits can also be distributed over multiple compute nodes.
 
-For mor information about running a mulitnode simulation, see [Multinode quantum
+For more information about running a mulitnode simulation, see [Multinode quantum
 simulation using HTCondor on Google Cloud](/qsim/tutorials/multinode).
 
 ## Runtime estimates
--- a/qsimcirq/qsim_circuit.py
+++ b/qsimcirq/qsim_circuit.py
@@ -343,7 +343,6 @@ class QSimCircuit(cirq.Circuit):
         cirq_circuit: cirq.Circuit,
         allow_decomposition: bool = False,
     ):
-
         if allow_decomposition:
             super().__init__()
             for moment in cirq_circuit:
@@ -352,6 +351,7 @@ class QSimCircuit(cirq.Circuit):
                     self.append(op)
         else:
             super().__init__(cirq_circuit)
+        self._check_for_confusion_matrix()
 
     def __eq__(self, other):
         if not isinstance(other, QSimCircuit):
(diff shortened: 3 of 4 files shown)


qsimcirq/qsim_simulator.py
```diff
@@ -213,19 +213,42 @@ def _run(
...
             A dictionary from measurement gate key to measurement
             results.
         """
+        self._check_for_confusion_matrix(circuit)
         param_resolver = param_resolver or cirq.ParamResolver({})
         solved_circuit = cirq.resolve_parameters(circuit, param_resolver)
 
         return self._sample_measure_results(solved_circuit, repetitions)
 
+    def _check_for_confusion_matrix(self, circuit: cirq.Circuit):
```
95-martin-orion, line 230: I recommend moving this to `qsim_circuit.py` and calling it in `QSimCircuit.translate_cirq_to_qsim` instead. Every Cirq circuit that gets converted to qsim must go through that method, so putting it there would save having to call it from every simulation method in this file.

qsimcirq/qsim_simulator.py
```diff
@@ -213,19 +213,42 @@ def _run(
...
+            op.gate.confusion_map
+            for _, op, _ in circuit.findall_operations_with_gate_type(
+                cirq.MeasurementGate
+            ) if op.gate.confusion_map
+        ]
+        for confusion_map in confusion_maps_on_measurement_gates:
+            for map_values in list(confusion_map.values()):
+                if map_values.size > 1:# Unsure if this check is needed. Do we simply need to see if *any*
+                    # confusion matrix exists at all or would a matrix of dimension (1x1) work?
+                    raise RuntimeError("Confusion Matrices are not currently supported in Qsim. "
```
95-martin-orion, line 249: Prefer `ValueError`, since this occurs when a value (`circuit`) is invalid.

qsimcirq/qsim_simulator.py
```diff
@@ -213,19 +213,42 @@ def _run(
...
+            Throws a runtime exception if a MeasurementGate with a confusion matrix is included in the circuit
+        """
+        confusion_maps_on_measurement_gates = [
+            op.gate.confusion_map
+            for _, op, _ in circuit.findall_operations_with_gate_type(
+                cirq.MeasurementGate
+            ) if op.gate.confusion_map
+        ]
+        for confusion_map in confusion_maps_on_measurement_gates:
+            for map_values in list(confusion_map.values()):
```
95-martin-orion, line 246: Drop `list` - you can iterate over the output of `values()` directly.

95-martin-orion (changes requested):

qsimcirq/qsim_simulator.py
angelo-laskaris: I moved the method to QSimCircuit and updated the unit test accordingly. However, I want to note the  invocation pattern seen here: https://github.com/quantumlib/Cirq/issues/6305 does not result in an exception being thrown.

angelo-laskaris:
Adding black formatter to my IDE, rebasing and then posting a commit on top of this one

qsimcirq/qsim_simulator.py
95-martin-orion: > the invocation pattern seen here: [quantumlib/Cirq#6305](https://github.com/quantumlib/Cirq/issues/6305) does not result in an exception being thrown.

That's deeply surprising to me. Could you try adding the check in `translate_cirq_to_qtrajectory` as well? https://github.com/quantumlib/qsim/blob/401ebec99628050d6e41cc43925027966f630ebd/qsimcirq/qsim_circuit.py#L438

angelo-laskaris: I added the check here as well and the `ValueError` still wasn't raised. Your surprise makes me want to make sure my environment doesn't have anything odd/wonky. Let me get a summary of my env (python version qsim ver, cirq ver) and follow up

angelo-laskaris: Python version: 3.9
-----
cirq==1.2.0
qsimcirq==0.16.3

angelo-laskaris:
@95-martin-orion The one test that failed doesn't seem to be related to the change. Is this test ever flaky?

95-martin-orion:
Opened #632 to track the issue.

95-martin-orion:
#632 is fixed - I'm going to go ahead with the v0.17.0 release to unblock some things, but I'm happy to cut a v0.17.1 once this is merged.

angelo-laskaris:
> #632 is fixed - I'm going to go ahead with the v0.17.0 release to unblock some things, but I'm happy to cut a v0.17.1 once this is merged.

Okay sounds good!

tanujkhattar:
@95-martin-orion Are there any further concerns or can we LGTM?

qsimcirq/qsim_simulator.py
95-martin-orion: How was this resolved?

95-martin-orion:
> @95-martin-orion Are there any further concerns or can we LGTM?

Sorry about the delay - this dropped off my radar. I've unresolved one comment above that I'd like clarity on, otherwise this is good to go. I'll follow up with a release shortly after.

qsimcirq/qsim_simulator.py
angelo-laskaris: I confirmed that there was nothing wonky with my setup or library versions and the indeed in neither `translate_cirq_to_qsim` or `translate_cirq_to_qtrajectory` when the circuit is executed like so

`qsimcirq.QSimSimulator(seed=0).run(circuit, repetitions=100)`

angelo-laskaris:
> > @95-martin-orion Are there any further concerns or can we LGTM?
> 
> Sorry about the delay - this dropped off my radar. I've unresolved one comment above that I'd like clarity on, otherwise this is good to go. I'll follow up with a release shortly after.

No worries! Followed up on your comment

95-martin-orion (approved):

PR #633, opened 2023-10-30
Pin CI python version

### LINKED ISSUE #632: Python test failing due to `distutils` deprecation
python 3.12 [no longer has a distutils package](https://stackoverflow.com/a/76691103), which causes issues in our test workflow. We either need to pin our python version to pre-3.12 or replace our usage of distutils.

Seen in #630 and #631.

95-martin-orion:
Fixes #632.

The python version for the `setup-python` action does not need to match the version used by `cibuildwheel`; it simply needs to be before `distutils` gets removed in python 3.12.

### DIFF
--- a/.github/workflows/release_wheels.yml
+++ b/.github/workflows/release_wheels.yml
@@ -45,6 +45,8 @@ jobs:
 
       # Used to host cibuildwheel
       - uses: actions/setup-python@v2
+        with:
+          python-version: '3.11'
 
       - name: Install cibuildwheel and twine
         run: python -m pip install cibuildwheel==2.12.3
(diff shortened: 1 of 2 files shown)


NoureldinYosri (approved):
Looks like qsim's dependency on distuils is two folds.
1. a superficial dependency in `setup.py`  to get `from distutils.version import LooseVersion` which should be replaced by `from packaging.version import parse` as per https://peps.python.org/pep-0632/#migration-advice
2. A transitive depdency that happens due to a dependency on an old version of tensorflow. 

So LGTM for now until we upgrade to a newer version of tf

PR #636, opened 2023-11-09
Update version to 0.17.1

95-martin-orion:
This release is for the confusion matrix guard added in #630.


95-martin-orion:
Between opening this and now #619 was merged, which is a signficant enough change to warrant a minor version bump instead. Shuffling PRs...

PR #640, opened 2023-11-15
Tighten boot disk lower bound

95-martin-orion:
The user who called out #639 narrowed the bound. They suspect as low as 32GB might be sufficient, but to allow files to grow somewhat before this needs another update I've opted to advertise 40GB as the new minimum.


NoureldinYosri (approved):

95-martin-orion:
Sanitizer tests are being flakier than usual - it's possible that one or more of the mirrors we use to find packages are down. I'll continue retrying for this PR, but we may need to consider finding new mirrors.

95-martin-orion:
Sanitizer flake was fixed in #641.

PR #643, opened 2023-11-20
Make qsimcirq Apple Silicon (M1, M2, ARM64) compatible and improve build scripts

basnijholt:
I've been working on compiling `qsimcirq` for `conda-forge` (https://github.com/conda-forge/staged-recipes/pull/24504) and encountered several challenges with the existing CMake scripts, particularly for Linux and MacOS x86 platforms.

This PR introduces modifications that ensure successful builds on these platforms. Additionally, I've extended support to Apple Silicon (ARM architecture), focusing on compiling `qsim_basic` due to the unavailability of most CPU instruction sets on ARM.

This inclusion is particularly beneficial for developers who use MacBook M1 laptops, as it enables them to write and test code on their local machines and seamlessly run the same code on more powerful computational clusters. 

Looking ahead, I plan to submit another PR to improve CUDA builds, further enhancing `qsimcirq`'s performance and compatibility.

I kindly request a review from @95-martin-orion, @jakurzak, and @sergeisakov, as your insights would be invaluable. Your feedback will help ensure that these changes align well with `qsim`'s development goals.

This PR aims to resolve the following issues: 
- Closes [#242](https://github.com/quantumlib/qsim/issues/242)
- Closes [#495](https://github.com/quantumlib/qsim/issues/495)
- Closes [#597](https://github.com/quantumlib/qsim/issues/597)

Thank you for considering these changes!

PR #646, opened 2023-12-04
Fix: Consistent Usage of `target_link_libraries` for CUDA Targets

basnijholt:
This pull request addresses a build configuration issue that arises (introduced in #643) when mixing plain and keyword signatures of `target_link_libraries` in CMake for the `qsim_cuda`, `qsim_custatevec`, `qsim_decide` targets. The existing setup led to a CMake error, as documented:

```
CMake Error at pybind_interface/cuda/CMakeLists.txt:48 (target_link_libraries):
  The plain signature for target_link_libraries has already been used with
  the target "qsim_cuda".  All uses of target_link_libraries with a target
  must be either all-keyword or all-plain.
```

### Changes Made:

- Modified the `target_link_libraries` command for the `qsim_cuda`, `qsim_custatevec`, `qsim_decide` targets to use the plain form. This change aligns with the internal usage of `target_link_libraries` within the `cuda_add_library` macro, ensuring consistency across the build configuration.

### Rationale:

CMake requires that all invocations of `target_link_libraries` for a specific target maintain a consistent usage pattern - either all using the plain form or all using the keyword form (such as `PUBLIC` or `PRIVATE`). The [`cuda_add_library` macro used in this project](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindCUDA.cmake#L2023-2031) employs the plain form of `target_link_libraries`, which necessitates a similar approach for additional invocations of `target_link_libraries` for the `qsim_cuda`, `qsim_custatevec`, `qsim_decide` targets.

By aligning the usage pattern, this pull request resolves the build configuration error and maintains consistency in the CMake setup, ensuring a smoother build process.

### DIFF
--- a/pybind_interface/cuda/CMakeLists.txt
+++ b/pybind_interface/cuda/CMakeLists.txt
@@ -27,4 +27,4 @@ set_target_properties(qsim_cuda PROPERTIES
 )
 set_source_files_properties(pybind_main_cuda.cpp PROPERTIES LANGUAGE CUDA)
 
-target_link_libraries(qsim_cuda PUBLIC OpenMP::OpenMP_CXX)
+target_link_libraries(qsim_cuda OpenMP::OpenMP_CXX)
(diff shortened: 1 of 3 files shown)


basnijholt:
@NoureldinYosri note that the CUDA build is currently broken and these changes are required.

PR #651, opened 2024-01-23
Drop dependency on distutils

NoureldinYosri:
The error https://github.com/quantumlib/qsim/issues/632 originally fixed in https://github.com/quantumlib/qsim/pull/633 has made a comeback https://fusion2.corp.google.com/invocations/f88845c0-24f7-4c02-b99d-74424ce80a75/targets/qsim%2Fmacos%2Fpresubmit/log.

Fix: Replace `distutils` with `packaging` as per https://peps.python.org/pep-0632/#migration-advice

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 set(CMAKE_CXX_STANDARD 11)
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.11)
 
 # Set APPLE_ARM to TRUE if running on Apple Silicon
 if(APPLE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 set(CMAKE_CXX_STANDARD 11)
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.12)
 
 # Set APPLE_ARM to TRUE if running on Apple Silicon
 if(APPLE)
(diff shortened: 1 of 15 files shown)


setup.py
```diff
@@ -27,7 +27,7 @@ def run(self):
             )
 
         if platform.system() == "Windows":
-            cmake_version = LooseVersion(
+            cmake_version = parse(
                 re.search(r"version\s*([\d.]+)", out.decode()).group(1)
             )
             if cmake_version < "3.1.0":
```
pavoljuhas, line 33: I think you need to parse the RHS as well here.

NoureldinYosri: what do you mean?

pavoljuhas, line 33:
```suggestion
            if cmake_version < parse("3.1.0"):
```

NoureldinYosri: done

CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 set(CMAKE_CXX_STANDARD 11)
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.12)
```
pavoljuhas, line 2: Is this needed?  Note that 3.12 here concerns the cmake version and has nothing to do with Python version.

NoureldinYosri: reverted

pavoljuhas:
Based on http://sponge2/200604ce-5698-4ab1-88be-403a0f5eff1c it seems we need to either upgrade the [tensorflow dependency](https://github.com/quantumlib/qsim/blob/9099e4eadfb9e5f2397e91c6aa51b27264802bfa/WORKSPACE#L19-L28) or somehow tell bazel to build with Python <= 3.11.

NoureldinYosri:
@pavoljuhas all checks pass, ptal

pavoljuhas (approved):
Nice, thanks for tracking this down!

Issue #655, opened 2024-02-09
missing py312 wheels


richrines1:
qsimcirq 0.20.0 adds support for python 3.12, however the corresponding wheels are [missing](https://pypi.org/project/qsimcirq/0.20.0/#files) from the pypi release so `pip install qsimcirq==0.20.0` still fails on python 3.12

NoureldinYosri:
thanks for reporting this. while we fix this you can install qsimcirq directly from github `pip install git+https://github.com/quantumlib/qsim`

PR #656, opened 2024-02-09
build wheel for 3.12

NoureldinYosri:
The version we are using of `cibuildwheel==2.12.3` is old and silently ignores py3.12 without raising an  error or warning this resulted in no wheel being released for 3.12 https://github.com/quantumlib/qsim/issues/655

fix: upgrade to `cibuildwheel==2.16.2`

---
before
before

after

after

---

Also I added 3.12 to windows as well since that was initially overlooked

---
fixes https://github.com/quantumlib/qsim/issues/655

### DIFF
--- a/.github/workflows/release_wheels.yml
+++ b/.github/workflows/release_wheels.yml
@@ -30,7 +30,7 @@ jobs:
             name: win_amd64
             architecture: x64
             cibw:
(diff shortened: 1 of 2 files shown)


95-martin-orion (approved):
LGTM! We've had this cibuildwheel problem a couple times before - glad to see it was relatively painless this time around.

PR #657, opened 2024-02-11
Update docs: Add option to use image with CUDA preinstalled

eliottrosenberg:
Describe the option to use a VM image that has CUDA preinstalled. Also redirect to a cached version of the cuQuantum Appliance website because the page is down.

### DIFF
--- a/docs/tutorials/gcp_gpu.md
+++ b/docs/tutorials/gcp_gpu.md
@@ -37,7 +37,7 @@ instance section, ensure that your VM has the following properties:
     3.   In the **Size** field, enter **40** (minimum).
 
     **Alternatively, you can click the "Switch Image" button and use the image with
-    CUDA pre-installed, which lets you skip step 3. I have verified that this works
+    CUDA pre-installed, which lets you skip step 3. It has been verified that this works
     with cuQuantum Appliance (option 3).**
 
 *   The instructions above override steps 3 through 5 in the [Create a Linux VM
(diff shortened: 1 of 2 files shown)


docs/tutorials/gcp_gpu.md
```diff
@@ -35,6 +35,11 @@ instance section, ensure that your VM has the following properties:
     1.   In the **Operating System** option, choose **Ubuntu**.
     2.   In the **Version** option, choose **20.04 LTS**.
     3.   In the **Size** field, enter **40** (minimum).
+
+    **Alternatively, you can click the "Switch Image" button and use the image with
+    CUDA pre-installed, which lets you skip step 3. I have verified that this works
```
sergeisakov, line 40: Could you perhaps replace "I have verified" with "It was verified" or something like that?

sergeisakov (changes requested):

docs/tutorials/gcp_gpu.md
```diff
@@ -176,8 +181,8 @@ to set up NVIDIA Container Toolkit.
 
 ## 9. Install NVIDIA cuQuantum Appliance (Option 3 only)
 Follow the instructions [here](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuquantum-appliance)
-to set up cuQuantum Appliance. You may need to use `sudo` for the Docker commands.
-
+to set up cuQuantum Appliance. If this link is broken, you can try the [cached page](https://webcache.googleusercontent.com/search?q=cache%3Ahttps%3A%2F%2Fcatalog.ngc.nvidia.com%2Forgs%2Fnvidia%2Fcontainers%2Fcuquantum-appliance&rlz=1CAFOSO_enUS1067&oq=cache%3Ahttps%3A%2F%2Fcatalog.ngc.nvidia.com%2Forgs%2Fnvidia%2Fcontainers%2Fcuquantum-appliance&gs_lcrp=EgZjaHJvbWUqBggAEEUYOzIGCAAQRRg7MgYIARBFGDoyBggCEEUYPDIGCAMQRRg80gEIMzMyMGowajSoAgCwAgA).
+You may need to use `sudo` for the Docker commands.
```
leofang, line 185: This is not appropriate, I'd say. If a link from NVIDIA is broken, what's expected is to report to NVIDIA (as you've kindly done - https://github.com/NVIDIA/cuQuantum/issues/118 - thank you!) and wait for resolution. Most likely it's transient and will be fixed shortly, unless it's because NVIDIA decides to discontinue the cuQuantum Appliance support, in which case we should nuke this section and look for alternatives. If we were to add a cache link to every single external page, it is unmaintainable. 

cc: @tlubowe @mtjrider for vis

docs/tutorials/gcp_gpu.md
eliottrosenberg: Done.

docs/tutorials/gcp_gpu.md
eliottrosenberg: I removed that sentence, thanks.

sergeisakov (approved):
LGTM

docs/tutorials/gcp_gpu.md
leofang: Thank you!

PR #665, opened 2024-07-02
Use bazel-contrib/setup-bazel instead of manually installing a deb during CI

dougthor42:
Use [`bazel-contrib/setup-bazel`](https://github.com/bazel-contrib/setup-bazel) instead of manually installing a deb during CI

Also fix some EOL whitespace because that's how my vim is configured.

@95-martin-orion

### DIFF
--- a/.github/workflows/bazeltest.yml
+++ b/.github/workflows/bazeltest.yml
@@ -22,12 +22,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - uses: bazel-contrib/setup-bazel@0.8.5
+        bazelisk-cache: true
+        disk-cache: ${{ github.workflow }}
+        repository-cache: true
       - name: Checkout submodules
         run: git submodule update --init --recursive
-      - name: Install Bazel on CI
-        run: |
-          wget https://github.com/bazelbuild/bazel/releases/download/5.3.0/bazel_5.3.0-linux-x86_64.deb
-          sudo dpkg -i bazel_5.3.0-linux-x86_64.deb
       - name: Install requirements
         run: |
           python3 -m pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/bazeltest.yml
```diff
@@ -22,12 +22,14 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - uses: bazel-contrib/setup-bazel@0.8.5
```
95-martin-orion, line 25: It's not immediately clear to me how (or if) this specifies the `bazel` version to install. IIRC, this is important for qsim as it needs to align with the bazel version used in [TFQ](https://github.com/tensorflow/quantum) for compatibility reasons.

Can we lock the bazel version to `5.3.0` here?

95-martin-orion (changes requested):

.github/workflows/bazeltest.yml
dougthor42: The bazel version is locked by `.bazelversion` in the root of the repo, and that's already set to `5.3.0`.

`setup-bazel` will install bazelisk 1.x as `bazel` and then any time `bazel` is called, bazelisk will download and install the bazel version defined by `.bazelversion` if the file exists (which it does :upside_down_face:)

More info on how bazelisk determines which version of bazel to run can be found here: https://github.com/bazelbuild/bazelisk?tab=readme-ov-file#how-does-bazelisk-know-which-bazel-version-to-run

dougthor42:
BTW, the context for this is https://github.com/qh-lab/pyle/pull/46833#issuecomment-2204296035

95-martin-orion (approved):
Changes LGTM, although it looks like we may need to claw back MacOS support before merging.

95-martin-orion:
@dougthor42 - #672 fixed the CI failures. If you're still interested in getting this merged, resolve conflicts and retry.

If I'm not available for final approval, request review from @pavoljuhas.

dougthor42:
Conflicts fixed, please approve CI run.

PR #678, opened 2024-12-27
Update dev-requirements.txt

95-martin-orion:
This should resolve https://github.com/quantumlib/qsim/security/dependabot/1.

### DIFF
--- a/.github/workflows/release_wheels.yml
+++ b/.github/workflows/release_wheels.yml
@@ -17,24 +17,24 @@ jobs:
             env:
               macosx_deployment_target: "12.0"
             cibw:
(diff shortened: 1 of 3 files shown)


95-martin-orion:
Also dropped python 3.9 since Cirq no longer supports it ([ref](https://github.com/quantumlib/Cirq/blob/686766fcc48178c7c4a54bc80e9d6c5ca9a139a0/.github/workflows/ci.yml#L171)).

95-martin-orion:
**Reviewers**: I'm too far removed from qsim at this point to be able to resolve the test failures effectively, but we should resolve the underlying vulnerability - if nothing else, it would stop Dependabot from spamming us every week 🙃

PR #699, opened 2025-05-11
Increment version number in master

mhucka:
The version number should have been incremented to 0.22.0.dev0 after the 0.21.0 release, but never was.

### DIFF
--- a/qsimcirq/_version.py
+++ b/qsimcirq/_version.py
@@ -1,3 +1,3 @@
 """The version number defined here is read automatically in setup.py."""
 
-__version__ = "0.21.0"
+__version__ = "0.22.0.dev0"


qsimcirq/_version.py
```diff
@@ -1,3 +1,3 @@
 """The version number defined here is read automatically in setup.py."""
 
-__version__ = "0.21.0"
+__version__ = "0.22.0.dev0"
```
dstrain115, line 3: Is it supposed to be 0.22.0.dev0 or 0.22.0.dev?

mhucka: The [PyPA guide to valid version numbers](https://packaging.python.org/en/latest/discussions/versioning/#valid-version-numbers) uses `x.y.z.devN`, and this is what Qualtran and Cirq are doing. See, for example, [Qualtran's version file](https://github.com/quantumlib/Qualtran/blob/main/qualtran/_version.py).

mhucka: (BTW, I know the CI checks are failing, and I'm also working on getting them running.)

Issue #705, opened 2025-05-23
build_tools/test_libs.sh is broken and cannot exit with an error


mhucka:
### Describe the issue

The test portion of `build_tools/test_libs.sh` cannot return a failure code – it's simply impossible the way it's written:

```bash
set e  # Ignore errors until artifacts are collected.
EXIT_CODE=0
for TARGET in bitstring_test channels_cirq_test circuit_qsim_parser_test expect_test \
              fuser_basic_test gates_qsim_test hybrid_avx_test matrix_test qtrajectory_avx_test \
              run_qsim_test run_qsimh_test simulator_basic_test simulator_sse_test statespace_basic_test \
              statespace_sse_test unitary_calculator_basic_test unitary_calculator_sse_test \
              unitaryspace_basic_test unitaryspace_sse_test vectorspace_test; do \
  if ! bazel test --test_output=errors tests:${TARGET}; then
    EXIT_CODE=1
  fi
done
```

First it sets the `e` Bash flag, which means errors in `bazel test` runs will not cause the script to exit. Then it sets and resets the variable `EXIT_CODE`, but the variable value is ultimately never used. If the script makes it through to this part of the file (which it always will on non-Linux machines), then it will always exit with success.

### What version of qsim are you using?

0.21

### How can the issue be reproduced?

Inspection of the script shows the problem.

mhucka:
Resolved by PR #710.

PR #710, opened 2025-05-23
Rewrite build_tools/test_libs.sh

mhucka:
This addresses issue #705, and also improves the script by making it more efficient and adding a --help option.

I removed the lines

```bash
which bazel
bazel version
```

because they never seemed useful in my uses.

### DIFF
--- a/build_tools/test_libs.sh
+++ b/build_tools/test_libs.sh
@@ -1,29 +1,47 @@
-#!/bin/bash
-set -e  # fail and exit on any command erroring
-set -x  # print evaluated commands
+#!/usr/bin/env bash
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
-which bazel
-bazel version
+set -eo pipefail -o errtrace
+shopt -s inherit_errexit
 
-# Attempt to build all components in SSE and basic mode.
-# The Github Action MacOS VMs may run on different-capability CPUs, so all AVX tests
-# are excluded from the build and test process.
-# apps are sample applications for experts and are meant to be run on only on Linux.
+declare -r usage="Usage: ${0##*/} [-h | --help | help] [bazel options ...]
+Run the programs in tests/, and on Linux, also build the programs in apps/.
 
-if [[ "$OSTYPE" == "linux-gnu"* ]]; then
-  bazel build --config=sse apps:all
-  bazel build apps:all
+If the first option on the command line is -h, --help, or help, this help text
+will be printed and the program will exit. Any other options on the command
+line are passed directly to Bazel.
+
+Note: the MacOS VMs in GitHub runners may run on different-capability CPUS, so
+all AVX versions of programs in tests/ are excluded."
+
+# Exit early if the user requested help.
+if [[ "$1" == "-h" || "$1" == "--help" || "$1" == "help" ]]; then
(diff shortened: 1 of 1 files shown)


mhucka:
Resolves issue #705.

dev_tools/test_libs.sh
```diff
@@ -1,29 +1,85 @@
...
+
+declare features=""
+shopt -s nocasematch
+# Note: can't use Bash $OSTYPE var here b/c the value is "linux-gnu" on Win 10.
+case "$(uname -s)" in
+    darwin*)
+        features=$(sysctl machdep.cpu.features)
+        ;;
+    linux*)
+        features=$(grep -si flags /proc/cpuinfo)
```
pavoljuhas: On my box this matches "flags" and "vmx flags" for each of 12 cores giving 24 lines in total.  I would also delete the `-s` flag.  It suppresses error message should /proc/cpuinfo not exist, but grep will still give failure status.  The script would exit on that line without any indication why.
Consider changing to

```suggestion
        features=$(grep -m1 -i "^flags" /proc/cpuinfo)
```

dev_tools/test_libs.sh
```diff
@@ -1,29 +1,85 @@
...
+if ! $avx || ! $sse; then
+    if ! $avx; then
+        filters+="-avx"
+        if ! $sse; then
+            filters+=",-sse"
+        fi
+    elif ! $sse; then
+        filters+="-sse"
+    fi
+fi
```
pavoljuhas: With `set -e` on line 16, the `[[ "$features" == *"avx2"* ]] && avx=true` will exit the script for non-avx2 processors, because the exit status of that command would be 1.  There is no need to have quotes in `*"avx2"*` as there is nothing expandable in  quoted string, `*avx2*` would do the same.  Also the `avx` and `sse` variables are effectively needed only once to build the filters string, so we can replace them with a conditional:
```suggestion
# Unless we can tell this system supports AVX, we skip those tests.
declare filters=""
[[ "$features" == *avx2* ]] || filters+=",-avx"
[[ "$features" == *sse* ]] || filters+=",-sse"
filters="${filters#,}"
```

dev_tools/test_libs.sh
```diff
@@ -1,29 +1,85 @@
...
+
+declare build_filters=""
+declare test_filters=""
+if [[ -n "$filters" ]]; then
+    build_filters="--build_tag_filters=$filters"
+    test_filters="--test_tag_filters=$filters"
+fi
+
+# Apps are sample programs and are only meant to run on Linux.
+# shellcheck disable=SC2086
```
pavoljuhas: The shellcheck directive can be avoided by using arrays for build and test filters instead of strings, consider applying

```diff
diff --git a/dev_tools/test_libs.sh b/dev_tools/test_libs.sh
index fd9cacc..415dcb0 100755
--- a/dev_tools/test_libs.sh
+++ b/dev_tools/test_libs.sh
@@ -67,19 +67,17 @@ if ! $avx || ! $sse; then
 fi
 
-declare build_filters=""
-declare test_filters=""
+declare -a build_filters=()
+declare -a test_filters=()
 if [[ -n "$filters" ]]; then
-    build_filters="--build_tag_filters=$filters"
-    test_filters="--test_tag_filters=$filters"
+    build_filters=( "--build_tag_filters=$filters" )
+    test_filters=( "--test_tag_filters=$filters" )
 fi
 
 # Apps are sample programs and are only meant to run on Linux.
-# shellcheck disable=SC2086
 if [[ "$OSTYPE" == "linux-gnu"* ]]; then
-    bazel build $build_filters --config=sse "$@" apps:all
-    bazel build $build_filters "$@" apps:all
+    bazel build "${build_filters[@]}" --config=sse "$@" apps:all
+    bazel build "${build_filters[@]}" "$@" apps:all
 fi
 
 # Run all basic tests. This should work on all platforms.
-# shellcheck disable=SC2086
-bazel test $build_filters $test_filters "$@" tests:all
+bazel test "${build_filters[@]}" "${test_filters[@]}" "$@" tests:all

```

dev_tools/test_libs.sh
mhucka: Excellent again. I especially like the trick of removing a leading comma on the last line, instead of trying to avoid it via the complicated logic. Much better.

Issue #695, opened 2025-05-09
Update version of Bazel


mhucka:
The Bazel configuration currently uses version 5.3.0, which is getting on in age. We should update to at least the version used in other Quantumlib projects (6.5.0).

mhucka:
Resolved by PR #711.

PR #711, opened 2025-05-23
Increase Bazel version to 6.5.0

mhucka:
This brings it into parity with the version of Bazel used by TensorFlow Quantum. (We could probably increase the version higher. We use 6.5.0 in other projects like TFQ, so it was easier to move to this version instead of going higher and risking some as-yet-unknown compatibility issue.)


mhucka:
Resolves #695.

Issue #707, opened 2025-05-23
CMake version needs to be updated


mhucka:
### Describe the issue

The current version specifiers in the various CMake files cause failures on newer versions of CMake. Updating the CMake files would resolve this problem.

### What version of qsim are you using?

0.21.0

mhucka:
Resolved by PR #714.

PR #714, opened 2025-05-23
Update min version specification in CMakeLists.txt

mhucka:
This increases the values supplied in `cmake_minimum_required` directives from the previous 3.11 to 3.31, so that CMake does not complain about the minimum version being too old.


mhucka:
Resolves issue #707.

dstrain115 (approved):

PR #724, opened 2025-05-28
Add pytest-xdist to dev-requirements

mhucka:
This allows for parallel `pytest` execution. 

(By itself, this addition to `dev-requirements.txt` does nothing; the purpose is to make it a separate commit so that the xdist features can be used in updated workflows and test scripts.)

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,3 +1,4 @@
 black~=25.1.0
 flynt~=1.0
 pytest
+pytest-xdist[psutil]
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,4 @@
 black~=25.1.0
 flynt~=1.0
 pytest
-pytest-xdist[psutil]
+pytest-xdist


mhucka:
> Consider if the `[psutil]` extra is truly needed. IIUC, it only comes into play for `--numprocesses=logical` option which may not get used at all.

Huh. I was going by the explanation in the [installation instructions](https://pytest-xdist.readthedocs.io/en/stable/#installation), which state:

> To use psutil for detection of the number of CPUs available, install the psutil extra:
> `pip install pytest-xdist[psutil]`

Looking at [the code](https://github.com/pytest-dev/pytest-xdist/blob/ebfcb99072aed3503fda5175245f28895a0204bb/src/xdist/plugin.py#L32), it looks like it defaults to using `psutil.cpu_count()` if another flag is not given; i.e., not purely for logical cpu counts.

However, the rest of the function around that line show that pytest tries a bunch of other approaches if it doesn't have psutil. Tthere doesn't seem to be a strong reason to include psutil after all, and I agree with you that it can be removed.

pavoljuhas (approved):

PR #763, opened 2025-06-12
Loosen NumPy version requirement to allow version 2

mhucka:
Set the min version of NumPy to 1.26 and allow NumPy 2.


requirements.txt
```diff
@@ -1,6 +1,6 @@
 absl-py
 cirq-core~=1.0
-numpy~=1.16
-pybind11
+numpy>=1.26.0
+pybind11[global]
```
dstrain115, line 6: What does the global signify?  Is that because we added platform dependent stuff in the other PRs?

dstrain115 (approved):

requirements.txt
mhucka: @dstrain115 The global option makes it include pybind11 CMake files in a location accessible from a more standard system location. In addition to going into $VIRTUAL_ENV/lib/python3.9/site-packages/pybind11/share/cmake, it goes into $VIRTUAL_ENV/share/cmake. This makes it possible for CMake to find them on some platforms (I think Windows was the problematic one).

dstrain115: Is it worth a one-line comment to explain this?

mhucka: That makes sense. I'll do that.

PR #782, opened 2025-06-18
Correct the default Bazel vector instructions options

mhucka:
The documented way that `bazel build`, `bazel test`, and similar commands behave without options like `--config=avx` is that the AVX, SSE, and other optimization options are not applied. However, this was not what was happening: the default ended up applying them even if `--config=sse` etc. were not used. Evidently, the previous changes to `.bazelrc` and `dev_tools/test_libs.sh` that I did in PR #758 and other recent PRs not correct this.

The changes here invert the way `--build_tag_filters` and `--test_tag_filters` are used: they are set to filter out things tagged with `avx` and `sse` by default in `.bazelrc`, and options like `--config=sse` add the things back.

To make this work, `dev_tools/test_libs.sh` has to be updated as well.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -23,17 +23,17 @@ common --verbose_failures
 # Windows uses slightly different syntax.
 common --enable_platform_specific_config
 
-build:linux --copt -std=c++17
-build:linux --cxxopt -std=c++17
+build:linux --copt=-std=c++17
+build:linux --cxxopt=-std=c++17
 
-build:macos --copt -std=c++17
-build:macos --cxxopt -std=c++17
+build:macos --copt=-std=c++17
+build:macos --cxxopt=-std=c++17
 
-build:windows --copt /std:c++17
-build:windows --cxxopt /std:c++17
+build:windows --copt=/std:c++17
+build:windows --cxxopt=/std:c++17
 
-build --copt -D_GLIBCXX_USE_CXX11_ABI=1
-build --cxxopt -D_GLIBCXX_USE_CXX11_ABI=1
+build --copt=-D_GLIBCXX_USE_CXX11_ABI=1
+build --cxxopt=-D_GLIBCXX_USE_CXX11_ABI=1
 
 # The default for vector instruction sets is to exclude them.
 # Config options later in this file can be layered to enable them.
@@ -68,13 +68,13 @@ build:sanitizer --copt -fno-omit-frame-pointer
 
 # Address sanitizer
 build:asan --config=sanitizer
-build:asan --copt -fsanitize=address
-build:asan --linkopt=-fsanitize=address
+build:asan --copt=-fsanitize=address
+build:asan --linkopt -fsanitize=address
 
 # Memory sanitizer
 build:msan --config=sanitizer
-build:msan --copt -fsanitize=leak
-build:msan --linkopt=-fsanitize=leak
+build:msan --copt=-fsanitize=leak
+build:msan --linkopt -fsanitize=leak
 
 # No sanitizers
 build:nosan --
@@ -103,7 +103,7 @@ build:basic --copt -O3
 
 # Build with OpenMP
 build:openmp --copt -fopenmp
-build:openmp --linkopt=-lgomp
+build:openmp --linkopt -lgomp
 
 # No OpenMP
 build:nopenmp --
@@ -112,7 +112,7 @@ build:nopenmp --
 # ~~~~ Memory handler (choose one) ~~~~
 
 # Build with tcmalloc
-build:tcmalloc --linkopt="-ltcmalloc"
+build:tcmalloc --linkopt -ltcmalloc
 
 # Build using malloc (default)
 build:malloc --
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -99,7 +112,7 @@ build:nopenmp --
 # ~~~~ Memory handler (choose one) ~~~~
 
 # Build with tcmalloc
-build:tcmalloc --linkopt="-ltcmalloc"
+build:tcmalloc --linkopt -ltcmalloc
```
pavoljuhas: Does this make any difference w/r to the previous format?

There are some `--copt arg` --> `--copt=arg` changes above, but this one goes the other way.

dev_tools/test_libs.sh
```diff
@@ -28,12 +28,12 @@ if [[ "$1" == "-h" || "$1" == "--help" || "$1" == "help" ]]; then
...
 
-# Unless we can tell this system supports AVX, we skip those tests.
+# Look for AVX and SSE in the processor's feature flags.
 declare features=""
 declare filters=""
 features="$(python -c 'import cpuinfo; print(" ".join(cpuinfo.get_cpu_info().get("flags", [])))')"
-[[ "$features" == *avx2* ]] || filters+=",-avx"
-[[ "$features" == *sse* ]] || filters+=",-sse"
+[[ "$features" == *avx2* ]] && filters+=",avx"
+[[ "$features" == *sse* ]] && filters+=",sse"
```
pavoljuhas, line 36: This runs into the same problem as in #710 - the shell has the `-e` option so the script will exit here if any of `avx2` or `sse` features were missing.  To have a positive condition, you need to rewrite this as
```sh
if [[ "$features" == *avx2* ]]; then
    filters+=",avx"
fi
...
```

dev_tools/test_libs.sh
```diff
@@ -43,10 +43,10 @@ if [[ -n "$filters" ]]; then
     test_filters=( "--test_tag_filters=$filters" )
 fi
 
-# Apps are sample programs and are only meant to run on Linux.
+# The apps are sample programs and are only meant to run on Linux.
 if [[ "$OSTYPE" == "linux-gnu"* ]]; then
-    bazel build "${build_filters[@]}" --config=sse "$@" apps:all
     bazel build "${build_filters[@]}" "$@" apps:all
+    bazel build "$@" apps:all
```
pavoljuhas: ~IIUC the bazel docs [here](https://bazel.build/docs/user-manual#build-tag-filters), we should remove line 48.  With `build_filters` the bazel will build only the targets matching one of the tags, but on the next line 49 they are in build scope again.  I am not sure why is then the first build command needed.~

(NVM after re-reading PR description, please consider adding a short comment above 48.  The comment below however might still apply)

Also, please remove `build_filters` in the last line 53.  Per the same doc link, they don't do anything for `bazel test`.

pavoljuhas (changes requested):
`test_libs.sh` may exit too early, please correct.

.bazelrc
mhucka: I think the answer is no, and the inconsistency came about due to some back-and-forth debugging of a run command that took arguments and seemed to get parsed incorrectly. Still, the safest approach seems to be to use `=`, so in the most recent commit, I changed the ones that used spaces to use `=`.

dev_tools/test_libs.sh
mhucka: So, I tested the following, and it does not exit early. Do you have an environment where it doesn't work this way?

```shell
 #!/usr/bin/env bash

 set -eo pipefail -o errtrace

 declare foo=""
 declare bar="original"

 foo="something"
 [[ "$foo" == *notthere* ]] && bar="new string"

 echo $bar
```

dev_tools/test_libs.sh
mhucka: Regarding that line: it was that way (two command invocations) in the original script before I touched it. To be honest, I'm not sure why. I would have removed line 49, but I wasn't sure if I was missing something essential.

Let's remove it because now we had 2 pairs of eyes who don't think it makes sense.

Re `build_filters`: good catch; thanks.

mhucka: Latest 2 commits do those two changes.

dev_tools/test_libs.sh
pavoljuhas: I stand corrected - per [bash set -e doc](https://www.gnu.org/software///bash/manual/bash.html#The-Set-Builtin) for the `&&` and `||` lists it is only the last command that counts for `errexit`.  I got confused, because on my Debian box it indeed exits early in an interactive session.  This happens because interactive bash runs some fancy prompt functions which try to restore the exit status of the last command and that triggers `errexit`.

With your snippet (and the same system bashrc) this can be reproduced with
```
bash -i <snippet.bash
```

TLDR - it is OK as is.

pavoljuhas (approved):

dev_tools/test_libs.sh
mhucka: Wow, that's unobvious. Thanks for the info!

PR #784, opened 2025-06-18
Add cache control options for manual debugging runs

mhucka:
I ran into a problem that turned out to be due to Bazel caches. To help future debugging situations, this adds an additional flag to `workflow_dispatch` manual workflow runs. The flag lets the user disable the Bazel caches.

### DIFF
--- a/.github/workflows/ci_build_library.yaml
+++ b/.github/workflows/ci_build_library.yaml
@@ -34,6 +34,10 @@ on:
         description: 'Run with debugging options'
         type: boolean
         default: true
+      bazel-cache:
+        description: 'Use Bazel caches'
+        type: boolean
+        default: true
 
 permissions: read-all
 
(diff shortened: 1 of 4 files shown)


pavoljuhas (commented):
Are these using GHA cache?  Note you can remove all caches using `gh cache delete --all` which effectively disables cache for the next run.

I am not sure misbehaving cache is common enough to be worth the extra complexity in workflows.

mhucka:
> Are these using GHA cache? Note you can remove all caches using `gh cache delete --all` which effectively disables cache for the next run.

Yes, indeed, and I use that too. However, when debugging why some mysterious failure is occurring, it's not obvious that it's a caching issue, so I prefer to test it in a single run than to delete all the caches. Deleting all the caches means all other workflow caches are affected. Doing it while other workflows are running can cause those workflows to fail, if the deletion happens when those workflow jobs get to the setup-bazel step (which has actually happened to me, incredibly enough, but admittedly it's a fluke of timing that is rare).

> I am not sure misbehaving cache is common enough to be worth the extra complexity in workflows.

Well, it's true it's not essential. I can use the `gh` command for now.

PR #785, opened 2025-06-18
Replace sprintf() call with snprintf()

mhucka:
clang on MacOS complains that sprintf() is insecure and should be replaced with snprintf().

### DIFF
--- a/pybind_interface/pybind_main.cpp
+++ b/pybind_interface/pybind_main.cpp
@@ -40,7 +40,7 @@ template <typename T>
 T parseOptions(const py::dict &options, const char *key) {
   if (!options.contains(key)) {
     char msg[100];
-    std::sprintf(msg, "Argument %s is not provided.\n", key);
+    std::snprintf(msg, sizeof(msg), "Argument %s is not provided.\n", key);
     throw std::invalid_argument(msg);
   }
   const auto &value = options[key];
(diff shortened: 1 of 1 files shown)


pybind_interface/pybind_main.cpp
```diff
@@ -40,7 +40,7 @@ template <typename T>
 T parseOptions(const py::dict &options, const char *key) {
   if (!options.contains(key)) {
     char msg[100];
-    std::sprintf(msg, "Argument %s is not provided.\n", key);
+    std::snprintf(msg, sizeof(msg), "Argument %s is not provided.\n", key);
```
pavoljuhas: we should use C++ types if clang needs us to fix it here:
```
  std::string msg = std::string("Argument ") + key + " is not provided.\n";
```

pavoljuhas (changes requested):
Let us switch to c++ if compiler makes us to do so.

pybind_interface/pybind_main.cpp
mhucka: You're right. I'll make the change.

mhucka: @pavoljuhas Done in the latest commit.

fdmalone (approved):

PR #821, opened 2025-06-28
Add PR size-labeling workflow

mhucka:
This PR size-labeling workflow and script came from Cirq. It makes it so that PRs are automatically labeled with an indication of their size, such as `size: S` or `size: L`.

### DIFF
--- a/.github/workflows/pr-labeler.yaml
+++ b/.github/workflows/pr-labeler.yaml
@@ -1,0 +1,79 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-labeler.yaml
```diff
@@ -0,0 +1,79 @@
...
+# as annotations on the workflow's run summary page. If labels don't seem to be
+# getting applied, check the run summary page for errors.
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+name: Pull request labeler
+run-name: >-
+  Label pull request ${{github.event.pull_request.number}} by ${{github.actor}}
+
+on:
+  pull_request:
```
pavoljuhas: This should be `pull_request_target` as in Cirq.  We tried with `pull_request` and it failed on permissions.  `pull_request` is also more dangerous, because it runs the workflow with the PR code.  `pull_request_target` runs the code in `master` which has been code-reviewed.

dev_tools/ci/size-labeler.sh
```diff
@@ -0,0 +1,213 @@
...
+        exit $exit_status
+    fi
+}
+
+function compute_changes() {
+    local -r pr="$1"
+
+    local change_info
+    local -r keys_filter='with_entries(select([.key] | inside(["changes", "filename"])))'
+    change_info="$(jq_stdin "map(${keys_filter})" <<<"$(api_call "pulls/${pr}/files")")"
```
pavoljuhas: Is this based on an older version of the Cirq size-labeler.sh?

I recall fixing these constructs (`cmd <<<"$(api_call ...)"`) which ignore the exit status of the api_call (and run `cmd` regardless) for cirq.

Please consider syncing this with the current Cirq script which we know that works.

pavoljuhas (changes requested):
We need to use `pull_request_target` as in Cirq.
Also the labeling script should be synced with its current Cirq version.

mhucka:
The latest commit updates the version of the script and the workflow.

PR #824, opened 2025-06-29
Improve hardware feature detection in CMake files and do further refactoring & enhancement

mhucka:
Changes:

* Previously, the CMake configuration was such that it would always build the AVX2, AVX512 & SSE2 Pybind11 modules without testing whether the current system supported those options. The changes here use CMake features to detect whether the architectural features are in fact available on the host computer, and only attempt to build the appropriate modules. The code for testing this is in a new `CheckCPU.cmake` file.

* Previously, each of 8 or so `pybind_interface/` subdirectories had `CMakeLists.txt` files that contained the same text for setting certain compilation flags. This PR removes the duplication in favor of putting the settings into the top-level `CMakeLists.txt` file.

* On MacOS, this adds more include and link directories (specifically those commonly used with Homebrew) for finding the LLVM and the OMP include files and libraries.

* Finally, this PR includes a bit of refactoring:

   - The introduction of a new directory, `dev_tools/cmake/`, for placing `.cmake` files included by other files. It moves `pybind_interface/GetPybind11.cmake` into this new location alongside the new `CheckCPU.cmake` file.
   - The reorganization of much of the top-level `CMakeLists.txt` file, to (hopefully) organize some of the code more logically.
   - The addition of more `message()` statements to help understand what's happening.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,12 +76,20 @@ elseif(APPLE AND NOT CMAKE_APPLE_SILICON_PROCESSOR)
     check_cpu_support("sse4_1" CPU_SUPPORTS_SSE4)
 endif()
 
+# Configure LTO for compilers that support it.
+include(CheckIPOSupported)
+check_ipo_supported(RESULT HAVE_LTO)
+
 # ~~~~~ Configure the build ~~~~~
 
 if(WIN32)
     add_compile_options(/O2)
 else()
     add_compile_options(-O3 -D_GLIBCXX_USE_CXX11_ABI=1)
+endif()
+
+if(HAVE_LTO)
+    set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
 endif()
 
 include(dev_tools/cmake/GetPybind11.cmake)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -54,6 +60,30 @@ endif()
...
 find_package(OpenMP REQUIRED)
 
+cmake_host_system_information(RESULT HAVE_SSE2 QUERY HAS_SSE2)
+
+if(WIN32)
+    check_cxx_compiler_flag("/arch:AVX2" HAVE_AVX2)
+    check_cxx_compiler_flag("/arch:AVX512" HAVE_AVX512)
+else()
+    check_cxx_compiler_flag("-mavx2" HAVE_AVX2)
+    check_cxx_compiler_flag("-mavx512f" HAVE_AVX512)
```
sergeisakov: Is it really a good way to detect whether the architectural features are in fact available on the host computer? `check_cxx_compiler_flag` checks whether the C++ compiler supports a given flag. It seems the g++ compiler supports the `-mavx512f` flag even if the host CPU does not support AVX512 instructions.

PR #829, opened 2025-07-01
Add a pip constraints file & constrain Pillow version

mhucka:
Pillow was updated to version 11.3.0 about 12 hours ago, and this is causing our builds on GitHub runners to fail because of some issue involving Pillow building wheels. Until that's resolved, this PR adds a constraints file that limits the version of Pillow installed by pip commands in the workflows.


mhucka:
Replaced with PR #830.

PR #848, opened 2025-07-31
Fix #847: missing range checks in statespace_*.h files

### LINKED ISSUE #847: CodeQL scans report missing range checks in lib/statespace_*.h files
### Describe the issue

CodeQL scans flagged lines like 353 in `lib/statespace_avx512.h`, reporting that the use of offset `m` should follow the range check:

```C++
     while (rs[m] < csum && m < num_samples) {
```

In more detail:

> The program contains an and-expression where the array access is defined before the range check. Consequently the array is accessed without any bounds checking. The range check does not protect the program from segmentation faults caused by attempts to read beyond the end of a buffer.

The same error exists in the following files:

* statespace_basic.h
* statespace_sse.h
* statespace_avx512.h
* statespace_avx.h

0.23.0.dev0

_No response_

mhucka:
CodeQL scans flagged lines like 353 in `lib/statespace_avx512.h`, reporting that the use of offset `m` should follow the range check:

```C++
     while (rs[m] < csum && m < num_samples) {
```

In more detail:

> The program contains an and-expression where the array access is defined before the range check. Consequently the array is accessed without any bounds checking. The range check does not protect the program from segmentation faults caused by attempts to read beyond the end of a buffer.

The same issue exists in the following files:

* statespace_basic.h
* statespace_sse.h
* statespace_avx512.h
* statespace_avx.h

### DIFF
--- a/lib/statespace_avx.h
+++ b/lib/statespace_avx.h
@@ -399,7 +399,7 @@ class StateSpaceAVX :
           double re = p[16 * k + j];
           double im = p[16 * k + 8 + j];
           csum += re * re + im * im;
-          while (rs[m] < csum && m < num_samples) {
+          while (m < num_samples && rs[m] < csum) {
             bitstrings.emplace_back(8 * k + j);
             ++m;
           }
--- a/lib/statespace_avx512.h
+++ b/lib/statespace_avx512.h
@@ -350,7 +350,7 @@ class StateSpaceAVX512 :
           double re = p[32 * k + j];
           double im = p[32 * k + 16 + j];
           csum += re * re + im * im;
-          while (rs[m] < csum && m < num_samples) {
+          while (m < num_samples && rs[m] < csum) {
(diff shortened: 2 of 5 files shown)


pavoljuhas (approved):
There seems to be one more such case:
```patch
diff --git a/lib/statespace_cuda_kernels.h b/lib/statespace_cuda_kernels.h
index b54ebca..0bc4ba7 100644
--- a/lib/statespace_cuda_kernels.h
+++ b/lib/statespace_cuda_kernels.h
@@ -318,5 +318,5 @@ __global__ void SampleKernel(unsigned num_blocks,
         FP3 im = state[l + warp_size];
         csum += re * re + im * im;
-        while (rs[m] < csum && m < num_samples) {
+        while (m < num_samples && rs[m] < csum) {
           bitstrings[m++] = k0 + k;
         }

```

Otherwise LGTM.

mhucka:
The latest commit has that fixed.

PR #857, opened 2025-08-10
Fix #855: don't call cirq.unitary() on non-unitary matrices

### LINKED ISSUE #855: NoiseMixture tests in qsimcirq_tests.py fail with Cirq 1.6.0
### Describe the issue

The CI tests fail against Cirq 1.6.0 and the prerelease versions of Cirq. Example run: https://github.com/quantumlib/qsim/actions/runs/16846849990/job/47727587531

The error concerns the `NoiseMixture` parametrized test case in `qsimcirq_tests/qsimcirq_test.py::test_multi_qubit_noise`. Something in this case triggers an exception in `cirq.unitary(…)`:

```
        if isinstance(val, np.ndarray):
            if not linalg.is_unitary(val):
>               raise ValueError("The provided numpy array is not unitary.")
E               ValueError: The provided numpy array is not unitary.

../../../../.pyenv/versions/3.12.5/envs/qsim-py312/lib/python3.12/site-packages/cirq/protocols/unitary_protocol.py:119: ValueError
```

This error does not occur with Cirq 1.5.0.

0.23.0.dev0

After installing `cirq-core` version 1.6.0 locally, the error can be reproduced by running `pytest`:

```shell
pytest qsimcirq_tests/qsimcirq_tests.py -k test_multi_qubit_noise
```

Doing nothing else but uninstalling `cirq-core` and then installing `cirq-core` version 1.5.0, and repeating the test, results in the tests passing.

mhucka:
In Cirq 1.6.0, the function [`unitary(...)`](https://github.com/quantumlib/Cirq/blob/fe0dc2187ca3269c178526e8ba41083fa1a467c9/cirq-core/cirq/protocols/unitary_protocol.py#L81) in `cirq-core/cirq/protocols/unitary_protocol.py` changed. Whereas previously, if the value passed in was a NumPy array, the array was returned without doing anything else, `unitary(...)` was changed in [Cirq PR #7536](https://github.com/quantumlib/Cirq/issues/7536) to test that the NumPy array really _is_ unitary and raise an exception if it isn't. This arguably corrected a fault in `unitary(...)` because with the previous definition, it would return non-unitary results if given a NumPy array that wasn't already unitary. However, one of the tests in qsim depended on this flaw: the test case knowingly invoked `unitary(...)` with a non-unitary matrix. This recently led to exceptions being raised in the qsim tests with Cirq 1.6.0 but not with prior Cirq versions.

The solution in this case turned out to be very simple: the call to `cirq.unitary(...)` on line 1060 was unnecessary because the previous version of `unitary(...)` simply returned the value untouched anyway. Removing the call kept the previous behavior (which was to use a non-unitary matrix in that situation) and works in both Cirq 1.5.0 and 1.6.0.

(An additional change to reformatting a few lines elsewhere was needed to satisfy the CI checks for formatting.)

### DIFF
--- a/qsimcirq_tests/qsimcirq_test.py
+++ b/qsimcirq_tests/qsimcirq_test.py
@@ -1057,7 +1057,7 @@ class NoiseMixture(NoiseChannel):
         super().__init__(*args, **kwargs)
 
     def _mixture_(self):
-        return [(prob, cirq.unitary(op)) for prob, op, in self._prob_op_pairs]
+        return [(prob, op) for prob, op, in self._prob_op_pairs]
 
 
 @pytest.mark.parametrize(
--- a/qsimcirq_tests/qsimcirq_test.py
+++ b/qsimcirq_tests/qsimcirq_test.py
@@ -2025,11 +2025,11 @@ def test_qsimcirq_identity_expectation_value():
     for w, pauli in objs:
         pauli = pauli[::-1]
         hamiltonian += float(w) * cirq.PauliString(
-            cirq.I(cirq.LineQubit(i))
-            if p == "I"
-            else cirq.Z(cirq.LineQubit(i))
-            if p == "Z"
-            else None
+            (
+                cirq.I(cirq.LineQubit(i))
+                if p == "I"
+                else cirq.Z(cirq.LineQubit(i)) if p == "Z" else None
+            )
             for i, p in enumerate(pauli)
         )
 
--- a/qsimcirq_tests/qsimcirq_test.py
+++ b/qsimcirq_tests/qsimcirq_test.py
@@ -1057,6 +1057,16 @@ class NoiseMixture(NoiseChannel):
         super().__init__(*args, **kwargs)
 
     def _mixture_(self):
+        # Cirq's mixture() function in mixture_protocol.py returns tuples of
+        # the form (probability, unitary peration). It does this by applying
+        # Cirq's unitary() function to the second elements of the tuples
+        # returned from here. Now, the values in self._prob_op_pairs will be
+        # tuples of the form (probability, NoiseStep). NoiseStep defines a
+        # _unitary_() method that simply returns the array as-is. Thus, when
+        # Cirq's mixture() function gets the value returned here and calls
+        # unitary() on those NoiseStep objects, the values unitary() returns
+        # will not actually be unitary. This is done knowingly. The nonunitary
+        # values are eventually normalized in test_multi_qubit_noise().
         return [(prob, op) for prob, op, in self._prob_op_pairs]
 
 
--- a/qsimcirq_tests/qsimcirq_test.py
+++ b/qsimcirq_tests/qsimcirq_test.py
@@ -1058,7 +1058,7 @@ class NoiseMixture(NoiseChannel):
 
     def _mixture_(self):
         # Cirq's mixture() function in mixture_protocol.py returns tuples of
-        # the form (probability, unitary peration). It does this by applying
+        # the form (probability, unitary operation). It does this by applying
         # Cirq's unitary() function to the second elements of the tuples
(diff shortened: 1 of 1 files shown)


mhucka:
Note: the docker build failure is due to an unrelated issue.

qsimcirq_tests/qsimcirq_test.py
```diff
@@ -1057,7 +1057,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def _mixture_(self):
-        return [(prob, cirq.unitary(op)) for prob, op, in self._prob_op_pairs]
+        return [(prob, op) for prob, op, in self._prob_op_pairs]
```
dstrain115, line 1070: Is _mixture a standard protocol for qsim or is this only used for this test?   This is changing the return type for this method.

dstrain115: Oh, I didn't realize op was already a np.array.

mhucka: Hmm. It's true that the 2nd element in the tuples in `self._prob_op_pairs` are actually `NoiseStep` objects. Lemme see if I can improve this.

mhucka: OK, with the way this all routed between different functions, I can't find a way to avoid changing this return type without rewriting `test_multi_qubit_noise()`. Given that this object class is only used in this file and only for that one function, that seems like it would be poor ROI.

I added a comment explaning what is happening. @dstrain115 can you check that it makes sense?

mhucka: > Oh, I didn't realize op was already a np.array.

Actuallly, I was wrong about that in my earlier comment (which I deleted too slowly to save you typing that reply -- sorry about that). At this point in the flow of things, it's not an ndarray. The whole flow is a bit more complicated, which hopefully the new comment clarifies.

qsimcirq_tests/qsimcirq_test.py
```diff
@@ -1057,7 +1057,17 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def _mixture_(self):
-        return [(prob, cirq.unitary(op)) for prob, op, in self._prob_op_pairs]
+        # Cirq's mixture() function in mixture_protocol.py returns tuples of
+        # the form (probability, unitary peration). It does this by applying
```
pavoljuhas: `peration` --> `operation`

pavoljuhas (approved):
LGTM with 1-char typo fixed

mhucka:
> @mhucka, @dstrain115 - should we remove the [is_unitary](https://github.com/quantumlib/Cirq/blob/fe0dc2187ca3269c178526e8ba41083fa1a467c9/cirq-core/cirq/protocols/unitary_protocol.py#L117-L119) introduced in [quantumlib/Cirq#7419](https://github.com/quantumlib/Cirq/pull/7419)? (or we could replace it with a check for a square matrix)
> 
> As it is `cirq.unitary(val)` returns `val._unitary_()` without any check, but raises ValueError when `val` is a non-unitary array. `cirq_unitary(cirq.unitary(val))` can thus fail when `cirq.unitary(val)` works.

By "remove `is_unitary` I assume you mean remove the addition of the checks using `linalg.is_unitary()`? (I can't find where it introduced a function `is_unitary`, but maybe I missed it.)

IMHO, removing the check for whether it is unitary would go against the purpose of `unitary(…)`. That seems wrong too. Instead, shouldn't `_strat_unitary_from_unitary(…)` check that the value it gets from `_unitary_` actually is unitary? (Similarly for the other `_strat*` methods.)

mhucka:
I opend https://github.com/quantumlib/Cirq/issues/7572 to follow-up on the issue that `cirq.unitary(…)` may return non-unitary values.

PR #867, opened 2025-08-16
Add workflow to check that files have copyright & license statements

mhucka:
This workflow uses [`addlicense`](https://github.com/google/addlicense) to check that files added in a PR contain the standard copyright and license statements. The workflow can be used as a required CI check.

The `addlicense` program doesn't provide an easy way to tell it which files to ignore. A feature of this workflow is that it overcomes this limitation by reading file paths to ignore from the file `.github/workflows/config/license-checker-ignore.txt`. This makes it easy to update the list of files to be ignored without having to edit the workflow itself.


mhucka:
Will do this differently.

PR #873, opened 2025-08-24
Expand the bug issue template & add a template for questions

mhucka:
People are filing bug reports without providing us basic info like the qsim version. This PR revises the issue templates in the following ways:

1. Adds a question about the qsim version number in the bug template
2. Adds a question about the computing environment in the bug template
3. Adds a new template for questions
4. Turns off the "blank issue" template option

After this change, there will be 4 templates: bug reports, feature requests, questions, and tasks.

Here is a screenshot of the revised bug template:

<div align="center">
screenshot-2025-08-24-1320
</div>


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1 +1 @@
-blank_issues_enabled: true
+blank_issues_enabled: false
```
dstrain115: Why no blank issues?   I think it's nice if you have something that doesn't fit in another category for some reason.

mhucka: There is an issue that someone filed when the previous forms were in place, and evidently they must have used the blank issue form, because the issue lacked almost any useful info or structure. That particular issue was a question, and the previous set of forms lacked one for questions, so I added one. After I did that, I couldn't think of something that didn't fit into the 4 categories that will exist after this PR is merged (i.e., bug, feature request, question, or task). That, in combination with the desire to steer inexperienced users away from using the blank form, made me think it was worth trying to limit issues to the 4 categories.

I don't feel very strongly about it. @dstrain115 can you think of issue types that don't fit in those 4 categories? If so, or if your past experience has been that blank issues are important to keep around, then we can keep them enabled.

mhucka: A relevant point of comparison here is to the issue types for Cirq. The ones for Cirq overlap: "design discussion" is basically "feature request", and "project health issue" is a "bug". I'm not really sure there is value in distinguishing those that finely, to be honest.

.github/ISSUE_TEMPLATE/1-bug-report.yml
```diff
@@ -5,33 +5,63 @@ body:
...
+      required: false
+
+  - type: input
+    attributes:
+      label: Tell us the version of qsim or qsimcirq (if relevant)
+      description: |-
+        For qsim software problems, we need to know the exact version you are
+        using. If you are using the Python package and are running within a
+        terminal emulator, then typing the following command in the shell will
+        print the version number:<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
```
pavoljuhas, line 46: Is it possible to use quoted textblock instead of nbsp-s here?  (ie, the one below comes with a copy button)
```
python3 -c 'import qsimcirq; print(qsimcirq.__version__)'
```

.github/ISSUE_TEMPLATE/config.yml
pavoljuhas: The [No Type Issues](https://github.com/quantumlib/qsim/issues?q=is%3Aissue%20state%3Aclosed%20no%3Atype) seems to be the most popular in the past issues.  They can capture cases which are not a clear cut at the time of reporting - e.g., if qsim crashed after installation it could be due to incorrect action on the user side (ie, Question - how to install correctly) or because of genuine bug in qsim.

pavoljuhas (approved):
Let us keep the blank issue in.  Otherwise LGTM with a small optional suggestion.

.github/ISSUE_TEMPLATE/1-bug-report.yml
mhucka: As you can imagine, that's the first thing I tried. Unfortunately, in issue forms, the copy button does not appear for code blocks:

<div align="center">
image
</div>

Furthermore, in dark mode, IMHO it's much harder to see where you should write the answer because of the styling – specifically, the white code text attracts visual attention while the input box is harder to notice:

<div align="center">
image
</div>

So, faced with these two facts, I looked for another solution and came up with the inline code + spacing approach you see in the final version.

.github/ISSUE_TEMPLATE/config.yml
mhucka: > The [No Type Issues](https://github.com/quantumlib/qsim/issues?q=sort%3Aupdated-desc+is%3Aissue+state%3Aclosed+no%3Atype) seems to be the most popular in the past issues.

The qsim repo did not have _any_ issue templates before I added the current ones earlier this year, which means the only issue type that people could select was the blank issue type. So of course, it looks like the most popular :-).

PR #879, opened 2025-08-25
Update quantum trajectories.

sergeisakov:


CMakeLists.txt
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required(VERSION 3.31)
+cmake_minimum_required(VERSION 3.28)
```
mhucka, line 15: <s>@sergeisakov Do you need the version reverted down, or was this only an accident? If it was deliberate, what is the environment where 3.31 does not work? (It happens that 3.31 is the minimum to avoid some warnings in other places. But if you ran into portability problems, then maybe we need to go with 3.28.)</s>

Never mind, I just read your email about the need for 3.28. OK, I'll try to see what happens with 3.28 in some other situations.

Makefile
```diff
@@ -171,9 +171,9 @@ run-tests tests: $(TESTS)
 
 .PHONY: check-cuquantum-root-set
 check-cuquantum-root-set:
-	@if [[ -z "$(CUQUANTUM_ROOT)" ]]; then \
-	    echo Error: '$$CUQUANTUM_ROOT must be set in order to use cuStateVec.' \
-	    exit 1 \
+	@if test -z "$(CUQUANTUM_ROOT)"; then \
```
mhucka, line 174: What is this change for portability reasons on another platform? I probably introduced the use of Bash `[[ … ]]` in other places, so if it causes problems, I will go through and change the others too.

mhucka: Never mind; I ran into the portability problem on Ubuntu too.

PR #885, opened 2025-09-11
Fix the shell script indentation -- should be 4, not 2

mhucka:
Changes:

* Fix the value for shell indentation in `.editorconfig`. It was incorrectly set to 2, and should be 4 instead.
* Add missing setting for makefiles in `.editorconfig`.
* In some other files, the relevant value of a linter configuration property should match the value of the editor configuration in `.editorconfig`. To help us remember this, I added appropriate comments to other config files.
* In reviewing the files, I now feel that the introductory comments at the tops of several files were too verbose and needlessly hard to maintain. I streamlined the comments.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -38,6 +38,9 @@ indent_size = 4
 [{CMakeLists.txt,*.cmake}]
 indent_size = 4
 
+[{Makefile,makefile,*.mk}]
+indent_size = 4
+
 [{*.cc,*.h}]
 indent_size = 2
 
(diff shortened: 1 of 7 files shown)


.editorconfig
```diff
@@ -41,6 +38,9 @@ indent_size = 4
 [{CMakeLists.txt,*.cmake}]
 indent_size = 4
 
+[{Makefile,makefile,*.mk}]
+indent_size = 4
```
pavoljuhas, line 43: Makefiles use TAB for indentation, please correct as in https://editorconfig.org/#example-file

Ref: https://stackoverflow.com/a/28720186/24247414

mhucka: Nice catch!

mhucka: Done in the latest commit.

PR #900, opened 2025-10-02
Change Dockerfiles to use cmake 3.28

mhucka:
This matches the recent changes in other files to use cmake 3.28.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,8 +27,8 @@ RUN apt-get update && \
 
 # Ubuntu 24's version of CMake is 3.28. We need a newer version.
 RUN apt-get remove --purge --auto-remove cmake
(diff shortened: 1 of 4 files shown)


fdmalone (approved):
Why was it rolled back?

mhucka:
> Why was it rolled back?

Sergei noted in his recent big PR that the default version of cmake on Ubuntu Linux is currently 3.28. Originally, I thought there was something in version 3.31 that was needed, but after testing with 3.28, I couldn't find anything breaking or even warnings, so whatever it was, it doesn't seem to have been critical.

(Some references to 3.31 in the qsim cmake files were already changed to 3.28, but I missed some others. This PR should catch the rest.)

PR #907, opened 2025-10-21
Fix problems in README.md and make minor updates

mhucka:
Changes:

* There was a link at the top to a section titled "features", but there was no such section. Fixed by rearranging and rewriting the lead-in text to split it between an intro paragraph and a "features" section.

* Remove mention of Conda. We currently don't have Conda builds.

* The input format is deprecated. It was previously suggested to me to remove mention of it from the README. Done.

* Removed the in-line BibTeX code in preference for using the approach we use in Cirq, which is to link to Zenodo.

* Fix miscellaneous markdownlint warnings.

### DIFF
--- a/.github/workflows/cirq_compatibility.yml
+++ b/.github/workflows/cirq_compatibility.yml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: 'Nightly Cirq compatibility test'
+name: 'Cirq dev compatibility test'
 run-name: Regularly test code against latest Cirq dev release
 
 on:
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -92,14 +88,6 @@ qsim, please refer to the
...
 
-### Input format
-
-> [!WARNING]
-> This format is deprecated, and no longer actively maintained.
-
-The circuit input format is described in the
-[docs](https://github.com/quantumlib/qsim/blob/main/docs/input_format.md).
-
 ### Sample circuits
```
sergeisakov: As you deleted the "Input format" section, you can also delete this section.

mhucka: Oh, yes, you're right. Thank you for catching that!

mhucka: Updated in the last push.

PR #911, opened 2025-10-24
Add `.pylintrc`, update `dev-requirements.txt`, and address some trivial Pylint warnings

mhucka:
This adds a `.pylintrc` file based on the one from Cirq, with a change to the line length (80 for this project) and removal of one of the checks because of a difference in conventions. It also adds `pylint` to `dev-requirements.txt` (using the same version that Cirq uses). Finally, it also adjusts a few Python files to address consequent linter warnings.

### DIFF
--- a/qsimcirq/qsim_simulator.py
+++ b/qsimcirq/qsim_simulator.py
@@ -514,8 +514,7 @@ class QSimSimulator(
         qubit_order: cirq.QubitOrderOrList = cirq.ops.QubitOrder.DEFAULT,
         initial_state: Any = None,
     ) -> Tuple[cirq.ParamResolver, np.ndarray, Sequence[int]]:
-        """Same as simulate() but returns raw simulation result without
-        wrapping it.
+        """Same as simulate() but returns raw simulation result without wrapping it.
 
         The returned result is not wrapped in a StateVectorTrialResult but can be used
         to create a StateVectorTrialResult.


.pylintrc
```diff
@@ -0,0 +1,51 @@
+[MAIN]
+load-plugins=pylint.extensions.docstyle,pylint.extensions.docparams
+max-line-length=80
```
pavoljuhas: 1. Please use 88 as in #912
2. The current incarnation ignores line length because there is no `line-too-long` in the enable list
3. The pylintrc in Cirq this is based on has been excised of rules that are checked by `ruff`.  If you do not plan on setting up ruff here, please use https://github.com/quantumlib/Cirq/blob/a025e0b1ab106cab1ebe2642ad42485a84b43467/dev_tools/conf/.pylintrc instead

dev-requirements.txt
```diff
@@ -1,6 +1,7 @@
 cmake~=3.28.1
 black~=25.9.0
 flynt~=1.0
+pylint~=3.3.1
```
pavoljuhas: As this seems to be an initial setup of pylint, let us use its most recent version

```suggestion
pylint~=4.0.2
```

setup.py
```diff
@@ -111,16 +109,18 @@ def build_extension(self, ext):
             ]
 
         env = os.environ.copy()
-        env["CXXFLAGS"] = '{} -DVERSION_INFO=\\"{}\\"'.format(
-            env.get("CXXFLAGS", ""), self.distribution.get_version()
+        env["CXXFLAGS"] = (
+            env.get("CXXFLAGS", "")
+            + f'-DVERSION_INFO=\\"{self.distribution.get_version()}\\"'
```
pavoljuhas: The new code dropped space between `$CXXFLAGS` from the environment and `-DVERSION_INFO...`, please fix.

Please replace with a single f-string (which can be broken across several lines if too long).

.pylintrc
mhucka: 1. (Face palm) Thanks for catching that.

3. Oh, I see. OK, I'll update it to use the Cirq .pylintrc before it was ruffed up.

W.r.t. ruff, we should do it at some point, but at this time there's too much else to do to tackle that. I'll open an issue to remind us to do it in the future.

dev-requirements.txt
mhucka: No objections; I was merely following what Cirq is using. But given that I was also using the wrong version of the `.pylintrc` file, maybe that's irrelevant. I'll try 4.0.2 and see what happens.

.pylintrc
mhucka: `.pylintrc` updated in the latest commit.

setup.py
mhucka: Done in an upcoming push.

qsimcirq/qsim_simulator.py
```diff
@@ -506,10 +514,11 @@ def simulate_into_1d_array(
         qubit_order: cirq.QubitOrderOrList = cirq.ops.QubitOrder.DEFAULT,
         initial_state: Any = None,
     ) -> Tuple[cirq.ParamResolver, np.ndarray, Sequence[int]]:
-        """Same as simulate() but returns raw simulation result without wrapping it.
+        """Same as simulate() but returns raw simulation result without
+        wrapping it.
```
pavoljuhas: Docstring summary should be a [single line](https://google.github.io/styleguide/pyguide.html#381-docstrings), please revert.  FWIW, I get no pylint complaints for the one line version on my side.

PR #912, opened 2025-10-24
Use same line length in .editorconfig as Black uses

mhucka:
`check/format-incremental` uses Black to check Python file formats, and it used the default Black configuration, which means line lengths are actually formatted to a max of 88 characters and not 80 as is used in other C++ files. So, `.editorconfig` needs to reflect this.


fdmalone:
don't some of our other repos use 100?

pavoljuhas (approved):

PR #918, opened 2025-10-29
Fix #884: test for `nvidia-smi` instead of `nvcc`

### LINKED ISSUE #884: The Makefile should test for nvidia-smi, not nvcc, to guess if CUDA available
### Describe the issue

It seems that checking for `nvidia-smi` is the more reliable way to determine if a system has a usable NVIDIA GPU and the necessary drivers installed for running CUDA applications. Checking for `nvcc` may only tell you if the system can compile CUDA code, not necessarily if it can run it.

### Tell us how to reproduce the issue

_No response_

### Tell us the version of qsim or qsimcirq (if relevant)

_No response_

### Tell us the computing environment (if relevant)

_No response_

mhucka:
It seems that checking for the program `nvidia-smi` is the more reliable way to determine if a system has a usable NVIDIA GPU and the necessary drivers installed for running CUDA applications. Checking for `nvcc` may only tell you if the system can compile CUDA code, not necessarily if it can run it.

This PR also fixes a small portability issue, where the use of `command -v` is preferrable to the use of `which` for testing whether a program is available.


Makefile
```diff
@@ -57,11 +57,14 @@ else
...
-
-ifneq (,$(shell which $(NVCC)))
+# CUDA if (i) we find nvidia-smi or (ii) $CUDA_PATH is set. For cuStateVec,
+# there's no way to find the cuQuantum libraries other than by being told, so
+# we rely on the user or calling environment to set variable $CUQUANTUM_ROOT.
+# (Testing for nvidia-smi is a more reliable way of checking that a system has
+# an NVIDIA GPU. Checking for nvcc may only tell you if the system can compile
+# CUDA code, not necessarily if it can run it.)
+
+ifneq (,$(shell command -v nvidia-smi > /dev/null 2>&1))
```
sergeisakov, line 67: There might be two issues here.

1. You discard both the standard output and the error output. So `ifneq (,$(shell command -v nvidia-smi > /dev/null 2>&1))` is always false.

2. It seems checking for `nvidia-smi` can also be unreliable. One can install `nvidia-utils` to get `nvidia-smi` (even on a machine that does not have an NVidia GPU) without installing cuda, etc. But we need `nvcc` to compile the code.

mhucka: > 1. You discard both the standard output and the error output. So `ifneq (,$(shell command -v nvidia-smi > /dev/null 2>&1))` is always false.

Ack, you're right of course. That should have been only `2> /dev/null`. And of course, I only bothered to test it on a system that didn't have nvidia-smi, so the resulting behavior was what I expected.  Anyway, thank you very for catching that.

> 2. It seems checking for `nvidia-smi` can also be unreliable. One can install `nvidia-utils` to get `nvidia-smi` (even on a machine that does not have an NVidia GPU) without installing cuda, etc. But we need `nvcc` to compile the code.

Darn it. I guess one would need to test for nvcc and nvidia-smi, then grep the output of running nvidia-smi for an indication of a gpu card. This is getting complicated :-(.

I'm doing to close this PR and issue #884 and try to reproduce whatever led me to think it was an issue in the first place.

PR #944, opened 2025-11-05
Rework CI workflow

mhucka:
Splitting up the previous workflows into separate files worked, but on balance, I find the disadvantages outweigh the advantages. The modularity of using separate files was somewhat good for reducing cognitive load and making maintenance a little bit easier, but it also decreased overall execution efficiency because it required more jobs overall and required more overhead.

This new single file `ci.yaml` puts all the CI checks into one place, and adds some additional lint jobs compared to the previous set of workflow files. It also introduces a separate action for setting up Bazel that lets us control more aspects of the configuration.

Note: this `ci.yaml` file does _not_ have the wheel-building steps that the previous setup had. The wheel builds and PyPI uploads will be reintroduced in an upcoming separate workflow.


.github/actions/set-up-bazel/action.yaml
```diff
@@ -0,0 +1,107 @@
...
+  steps:
+    - name: Determine the version of Bazel to use
+      shell: bash
+      env:
+        SHELLOPTS: ${{(inputs.debug || runner.debug) && 'xtrace' || ''}}
+      run: |
+        if [[ -n "${{inputs.bazel-version}}" ]]; then
+          version="${{inputs.bazel-version}}"
+        elif [[ -f ".bazelversion" ]]; then
+          version="$(cat .bazelversion | tr -d ' \n\r')"
```
pavoljuhas: nit - no need for cat:
```suggestion
          version="$(tr -d ' \n\r' <.bazelversion)"
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,522 @@
...
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+
+      - name: Get a list of changed Python files
+        id: changes
+        uses: tj-actions/changed-files@24d32ffd492484c1d75e0c0b894501ddb9d30d62 # v47
+        with:
+          base_sha: ${{inputs.base_sha}}
+          files: '**/*.py'
```
pavoljuhas: There are only 10 Python files in the repo, so the time required to run the format and lint check would be very short.  In addition the outcome of these checks can change after modifying their configuration or switching to newer versions.  As it is such changes may show up only at later PRs that touch some `.py` files.

I'd suggest to remove this step and any related if-logic below.

Also consider doing the same for other steps below that complete within few seconds.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,522 @@
...
+        run: docker run --rm qsim-base:latest -c /qsim/circuits/circuit_q24
+
+      - name: Test installation process
+        run: |
+          cd install/tests
+          docker compose build
+
+  report-results:
+    name: CI
+    if: 'always()'
```
pavoljuhas: should this be quoted?

pybind_interface/avx2/CMakeLists.txt
```diff
@@ -15,11 +15,14 @@
...
-IF (WIN32)
-    set(CMAKE_CXX_FLAGS "/arch:AVX2 /O2 /openmp")
-ELSE()
-    set(CMAKE_CXX_FLAGS "-mavx2 -mfma -O3 -flto=auto")
-ENDIF()
+if(WIN32)
+    add_compile_options(/arch:AVX2 /openmp)
+    # Add /O2 to any configuration that is NOT Debug.
+    # This prevents a conflict with /RTC1 in DEBUG builds.
+    add_compile_options($<$<NOT:$<CONFIG:Debug>>:/O2>)
```
pavoljuhas, line 22: I am not familiar with cmake syntax, but could this be put to a nested if statement which might be more readable?

(Same for other CMakeList.txt files in this PR).

.github/workflows/ci.yaml
mhucka: Both quoted and unquoted forms are valid, but you're right, the quotes are superfluous here. Removed in the next push.

.github/workflows/ci.yaml
mhucka: So, just so I'm clear, you mean to have the formatting checks run always, and not only on files that have been changed?

pybind_interface/avx2/CMakeLists.txt
mhucka: > I am not familiar with cmake syntax, but could this be put to a nested if statement which might be more readable?
> 
> (Same for other CMakeList.txt files in this PR).

Yeah, I agree that syntax is gnarly (and I confess I needed Gemini's help to come up with it). The thing is that's it's a build-time conditional, whereas an alternative like

```cmake
if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
    add_compile_options(/O2)
endif()
```

only works at configuration time (i.e., by re-running cmake), and apparently for MSVC and Xcode, `CMAKE_BUILD_TYPE` may be empty at configuration time. And there doesn't seem to be a build-time equivalent for the if-then syntax.

.github/workflows/ci.yaml
mhucka: OK, I'll take out the conditional logic. It's true that there are not a lot of files and the changed-file-detection is probably overcomplicating things.

pavoljuhas: > So, just so I'm clear, you mean to have the formatting checks run always, and not only on files that have been changed?

Yes, I think it is better to keep things simple and more robust, especially if checks are fast and cheap.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,491 @@
...
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends -y yamllint
+
+      - name: Lint the YAML files
+        continue-on-error: ${{env.soft-linting == 'true'}}
+        run: |
+          echo "::add-matcher::.github/problem-matchers/yamllint.json"
+          find . -name '*.yaml' -o -name '*.yml' | \
+            grep -vE '\\.github/workflows' | \
+            xargs yamllint -f github
```
pavoljuhas: The grep pattern requires a literal `\` so it cannot exclude `./.github/workflows`.
How about
```suggestion
          find . -not -path '*.github/workflows/*' \
            '(' -name '*.yaml' -o -name '*.yml' ')' | \
            xargs yamllint -f github
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,491 @@
...
+          echo ":white_check_mark: All CI checks passed" >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Point out if soft-linting is in effect
+        if: env.soft-linting
+        run: |
+          {
+          echo "> [!CAUTION]:"
+          echo "> **Soft linting is in effect**."
+          echo "> Format and lint errors have been ignored."
+          } >> "$GITHUB_STEP_SUMMARY"
```
pavoljuhas, line 488: I was trying to suggest soft checks as a temporary workaround to be removed as soon as possible so the workflow code is simple.  Is that the plan here?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,491 @@
...
+        uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v5
+        with:
+          python-version: ${{env.python-version}}
+          cache: pip
+          cache-dependency-path: |
+            requirements.txt
+            dev-requirements.txt
+
+      - name: Install dependencies
+        run: pip install black~=25.9.0 flynt~=1.0 pylint~=4.0.0
```
pavoljuhas: Please install `-r dev-requirements.txt` instead so you don't get out-of-sync versions here.  The python environment is cached just above so installing a few extra dependencies from dev-requirements should not matter.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,491 @@
...
+
+      - name: Test installation process
+        run: |
+          cd install/tests
+          docker compose build
+
+  report-results:
+    name: CI
+    if: always()
+    needs:
```
pavoljuhas, line 455: Is this needed?  I was under impression the default GHA handling of workflow status does essentially the same without extra code to maintain on our side.

mhucka: The purpose of this is to make it possible to add only the single "CI" job in the required checks, rather than all the individual jobs. Managing the individual jobs has been an error-prone chore, and this is a great approach to avoid it altogether. (To give credit where it's due: I got this idea from the Gemini CLI repo's approach.)

.github/workflows/ci.yaml
mhucka: Right, I understood it that way, but then

1) Ran into the problem that some of the linters run as actions that don't have a flag or a direct way to tell them not to error out (and since they're not invoked via a shell command, the error can't be masked somehow) so I had to add the field `continue-on-error:` to the step.

2) That led to the realization that a uniform way of handling soft linting is to put `continue-on-error` for all the relevant steps. It means maintainers only have to remember one pattern, instead of looking for individual flags or thinking about where to put a `|| true`. I think this is actually a simplification over what would have been done otherwise.

3) And then that led to the realization that there's value in having a real toggle during testing, so I added a parameter to workflow_dispatch. For example, if one wanted to test some of the later jobs even if a linter or formatter complained about something, this makes it easy. (This works both on GitHub and when running the workflow locally, by passing `--input soft-linting=true` to `gh act …`.)

If it's too much to have that final message-printing step, it could be removed.

.github/workflows/ci.yaml
mhucka: The purpose of this particular job is to make it possible to add only this single "CI" job in the required checks for GitHub pull requests, rather than all the individual jobs. Managing the individual jobs has been an error-prone chore, and this is a great approach to avoid it altogether. (To give credit where it's due: I got this idea from the Gemini CLI repo's approach.)

.github/workflows/ci.yaml
mhucka: That's how I started, but then

1) Ran into the problem that some of the linters run as actions that don't have a flag or a direct way to tell them not to error out (and since they're not invoked via a shell command, the error can't be masked somehow) so I had to add the field `continue-on-error:` to their step.

2) That led to the realization that a uniform way of handling soft linting throughout the workflow is to put `continue-on-error` for all the relevant steps. It means maintainers only have to remember one pattern, instead of looking for individual flags or thinking about where to put a `|| true`. I think this is actually a simplification over what would have been done otherwise.

3) And then that led to the realization that there's value in having a real toggle during testing, so I added a parameter to workflow_dispatch. For example, if one wanted to test some of the non-lint/format jobs even if a linter or formatter complained about something, this makes it easy. (This works both on GitHub and when running the workflow locally, by passing `--input soft-linting=true` to `gh act …`.)

If it's too much to have that final message-printing step, it could be removed.

PR #959, opened 2025-11-18
Fix #839,  move `setup.py` metadata & config settings into `pyproject.toml`, & remove `requirements.txt`

### LINKED ISSUE #839: Fix use of deprecated setuptools functions
### Describe the issue

Building qsim produces the following warning messages due to constructs used in setup.py:

```
/Users/mhucka/.local/lib/python3.10/site-packages/setuptools/__init__.py:92: _DeprecatedInstaller: setuptools.installer and fetch_build_eggs are deprecated.
!!

        ********************************************************************************
        Requirements should be satisfied by a PEP 517 installer.
        If you are using pip, you can try `pip install --use-pep517`.

        By 2025-Oct-31, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.
        ********************************************************************************

!!
  dist.fetch_build_eggs(dist.setup_requires)
```

We should update this to use whatever the modern setuptools equivalents are.

0.23.0.dev0

_No response_

mhucka:
This addresses deprecation warnings originating from the use of outdated `setuptools` functions, and takes the opportunity to modernize the management of the version number and the dependencies in accordance with current Python practices (PEP [621](https://peps.python.org/pep-0631/), [735](https://peps.python.org/pep-0735/)).

Key changes include:

- All declarative package metadata (e.g., name, author, classifiers) has been moved from `setup.py` to `pyproject.toml`. What's left in `setup.py` has been simplified to only contain the logic necessary for building the C++ extensions.

- The qsim version number is now stored in only on file, `qsimcirq/_version.py`, and both `pyproject.toml` and `setup.py` read it from there. In the case of `setup.py`, the use of the deprecated `self.distribution.get_version()` has been replaced with the use of `runpy` to read the version number from `qsimcirq/_version.py`. (The latter change fixes #839.)

- Development dependencies are no longer stored as "extras" but rather use the pyproject.toml  `[dependency-groups]` section introduced in 2024 by [PEP 735](https://peps.python.org/pep-0735/) and recognized by [pip version 25.1+](https://packaging.python.org/en/latest/specifications/dependency-groups/#specification). This means both `requirements.txt` and `dev-requirements.txt` are gone, and development dependencies are installed using
	```shell
	pip install --group dev
	```

- `pyproject.toml` defines the versions of Python for which cibuildwheel builds wheels (in section `[tool.cibuildwheel]`).

### DIFF
--- a/docs/install_qsimcirq.md
+++ b/docs/install_qsimcirq.md
@@ -15,10 +15,11 @@ Prior to installation, consider opening a
 
 Prerequisites are defined in the
 [`pyproject.toml`](https://github.com/quantumlib/qsim/blob/main/pyproject.toml)
-file, and will be automatically installed along with qsimcirq.
+file, and will be automatically installed along with qsimcirq when you install
+it with pip.
 
 If you'd like to develop qsimcirq, a separate set of dependencies for
-development are includes in the
+development are included in the
 [`pyproject.toml`](https://github.com/quantumlib/qsim/blob/main/pyproject.toml)
 file. You can install them with `pip3 install --group dev`. (You will need
 pip version 25.1 or higher.)


setup.py
```diff
@@ -110,9 +113,7 @@ def build_extension(self, ext):
 
         env = os.environ.copy()
         cxxflags = env.get("CXXFLAGS", "")
-        env["CXXFLAGS"] = (
-            f'{cxxflags} -DVERSION_INFO=\\"{self.distribution.get_version()}\\"'
-        )
+        env["CXXFLAGS"] = f'{cxxflags} -DVERSION_INFO=\\"{__version__}\\"'
```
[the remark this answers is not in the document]
mhucka: It is a goal to avoid doing this, because I'm working towards externalizing the version number out of setup.py.

pyproject.toml
```diff
@@ -13,14 +13,100 @@
...
+    "schrödinger-feynman simulation",
+    "sdk",
+    "simulation",
+    "state vector simulator",
+    "software development kit",
+]
+dependencies = [
+    "absl-py",
+    "cirq-core~=1.0",
+    "numpy>=1.26.0",
```
pavoljuhas: This duplicates `requirements.txt` - do you plan on retiring that file together with `dev-requirements.txt` and only declare dependencies in pyproject.toml?

You might want to check if qsim can be converted to a [uv](https://docs.astral.sh/uv/) managed project which will support dependency-tree locking and a full development reproducibility.

setup.py
```diff
@@ -24,6 +23,8 @@
 from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext
 
+__version__ = "0.23.0.dev0"
```
pavoljuhas: You can use [tomllib](https://docs.python.org/3.11/library/tomllib.html#module-tomllib) to get version from pyproject.toml and avoid duplicate definition.  That modules is builtin since Python 3.11

pyproject.toml
mhucka: Re duplication: you're right. Maybe it would be better to remove the other two files in this PR. I went back and forth about whether to do that in this PR, and now I think it would be better to make that change in a separate PR.

Re uv: it's a good idea, and we should do that at some point, but IMHO it would be better to leave that to a future qsim release in order to limit the differences in the next release (0.23). I opened issue https://github.com/quantumlib/qsim/issues/977 to remind us.

setup.py
mhucka: That's an excellent idea, although qsim is currently supposed to support 3.10. I guess it could be added to the installation dependencies with a condition on the python version. I'll try that.

Dockerfile
```diff
@@ -37,19 +38,17 @@
...
 # Activate venv.
 ENV PATH="/test_env/bin:$PATH"
 
-# Install qsim requirements.
-# hadolint ignore=DL3042
-RUN python3 -m pip install -r /qsim/requirements.txt && \
-    python3 -m pip install -r /qsim/dev-requirements.txt
+# Install qsim development requirements.
+# hadolint ignore=DL3013
+RUN python3 -m pip install --no-cache-dir --group dev
```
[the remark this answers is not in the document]
mhucka: This is deliberate.

setup.py
mhucka: The latest commits remove the duplication of version numbers and uses a modern way to get the version number in pyproject.toml.

PR #968, opened 2025-12-30
Fix #904: test that repo owner is Quantumlib

### LINKED ISSUE #904: Disable GitHub actions jobs in forks
### What is the task?

I noticed that my fork of the qsim repo executed GitHub actions after updating its main branch.  I did not change any GHA-related settings after creating the fork.

The qsim repo should have if-guard to skip execution in downstream forks so
users do not spend their GHA quota on qsim workflows.

Ref: https://github.com/quantumlib/Cirq/pull/7326

### (Optional) Do you have ideas or preferences for the approach?

_No response_

### How urgent is this for you?

None

mhucka:
This adds checks to the GitHub Actions workflows for the owner being Quantumlib, so that users who fork the repo will not be surprised by the workflows running in their forks.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,14 +56,6 @@ concurrency:
 permissions: read-all
 
 jobs:
-  check-owner:
-    if: github.repository_owner == 'quantumlib'
-    name: 'Check that the repository owner is Quantumlib'
-    runs-on: ubuntu-slim
-    timeout-minutes: 5
-    steps:
-      - run: true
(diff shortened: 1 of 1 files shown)


pavoljuhas (changes requested):
I feel we should use `if: github.repository_owner == 'quantumlib'` everywhere for the sake of uniformity in all workflows.

Also when running for the main repo the CI job will know the outcome instantaneously and would not need to wait for a container to spin up and run the `true` command.

PR #976, opened 2025-12-30
Add configuration for `check-wheel-contents`

mhucka:
The utility `check-wheel-contents` is useful for checking that wheels are constructed correctly; however, it generates false positives because it doesn’t know about how `cibuildwheel` adds object files to wheels.


mhucka:
After further thought, closing this PR. This should only be added if/after check-wheel-contents is used in a script or workflow.

PR #985, opened 2026-01-02
Move `setup.py` metadata into `pyproject.toml`, remove `dev-requirements.txt`, & fix #839

### LINKED ISSUE #839: Fix use of deprecated setuptools functions

mhucka:
This addresses deprecation warnings originating from the use of outdated `setuptools` functions, and takes the opportunity to modernize the management of the version number and the dependencies in accordance with current Python practices (PEP [621](https://peps.python.org/pep-0631/), [735](https://peps.python.org/pep-0735/)).

Key changes include:

- All declarative package metadata (e.g., name, author, classifiers) has been moved from `setup.py` to `pyproject.toml`. What's left in `setup.py` has been simplified to only contain the logic necessary for building the C++ extensions and calling `setup`.

- The qsim version number is now stored in only one file, `qsimcirq/_version.py`, and both `pyproject.toml` and `setup.py` read it from there. In the case of `setup.py`, the use of the deprecated `self.distribution.get_version()` has been replaced with the use of `runpy` to read the version number from `qsimcirq/_version.py`. (The latter change fixes #839.)

- The development dependencies are no longer stored as "extras" but rather use the pyproject.toml  `[dependency-groups]` section introduced in 2024 by [PEP 735](https://peps.python.org/pep-0735/) and recognized by [pip version 25.1+](https://packaging.python.org/en/latest/specifications/dependency-groups/#specification). This means `dev-requirements.txt` is gone, and development dependencies are installed using
	```shell
	pip install --group dev
	```

- `pyproject.toml` defines the versions of Python for which cibuildwheel builds wheels (in section `[tool.cibuildwheel]`).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,10 +3,10 @@
 # Core dependencies:
 absl-py
 cirq-core~=1.0
-numpy>=1.26.0,<2.0; python_version < '3.11'
-numpy>=2.0; python_version >= '3.11'
 
-# These are transitive dependencies we need to constrain to avoid unresolvable
-# installation conflicts due to them requiring higher Python versions:
-scipy<1.16; python_version < '3.11'
-contourpy<1.3; python_version < '3.11'
+# Constrain some versions because they dropped support for Python 3.10.
+# Countourpy is pulled in by Matplotlib, which is pulled in by Cirq.
+contourpy<1.3.0; python_version <= '3.10'
+numpy<2.0.0; python_version <= '3.10'
+numpy>=1.26.0; python_version > '3.10'
+scipy<1.15.0; python_version <= '3.10'
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -2,9 +2,9 @@
 
 This style guide for Gemini outlines the coding conventions for this project.
 
-## General guidance
+## Introduction and goals
 
-### Overall principles
+### Overall goals
 
 *   _Readability_: Code must be easy for humans to understand. Prefer clarity
     over cleverness. Write elegant, well-structured code.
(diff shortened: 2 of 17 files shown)


.github/workflows/ci.yaml
```diff
@@ -237,10 +242,15 @@
           cache: pip
           cache-dependency-path: |
             requirements.txt
-            dev-requirements.txt
+            pyproject.toml
+
```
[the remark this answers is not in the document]
mhucka: Implications understood.

.github/workflows/ci.yaml
```diff
@@ -237,10 +242,15 @@
...
+      - name: Upgrade pip
```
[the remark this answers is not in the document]
mhucka: Implications understood.

.github/workflows/cirq_compatibility.yml
```diff
@@ -52,15 +52,15 @@
...
+            pyproject.toml
 
       - name: Install latest dev version of Cirq
         run: pip install --upgrade cirq~=1.0.dev
 
       - name: Install qsim dev requirements
         run: |
           pip install -r requirements.txt
-          pip install -r dev-requirements.txt
+          pip install --group dev
```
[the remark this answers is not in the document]
mhucka: Implications understood.

pybind_interface/Dockerfile
```diff
@@ -26,5 +26,9 @@
 # Build pybind code early to cache the results
 RUN make -j -C /qsim/ pybind
 
+# Install Python development dependencies.
+# hadolint ignore=DL3013
+RUN pip install --no-cache-dir --group dev
```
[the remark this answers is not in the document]
mhucka: Implications understood.

Dockerfile
```diff
@@ -37,22 +37,24 @@
...
 # Activate venv.
 ENV PATH="/test_env/bin:$PATH"
 
 # Install qsim requirements.
-# hadolint ignore=DL3042
-RUN python3 -m pip install -r /qsim/requirements.txt && \
-    python3 -m pip install -r /qsim/dev-requirements.txt
+# hadolint ignore=DL3013
+RUN python3 -m pip install --no-cache-dir --upgrade pip && \
+    python3 -m pip install --no-cache-dir -r requirements.txt
```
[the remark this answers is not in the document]
mhucka: Implications understood.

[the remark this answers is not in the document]
mhucka: Implications understood.

.github/workflows/ci.yaml
```diff
@@ -73,10 +73,15 @@ jobs:
...
+        run: python -m pip install --upgrade pip
 
       - name: Install dependencies
-        run: pip install -r requirements.txt -r dev-requirements.txt
+        run: |
+          pip install -r requirements.txt
+          pip install --group dev
```
pavoljuhas: I think it is better to run these together in one pip process.
That way any shared dependencies would need to resolve consistently for both requirements.txt and the dev group.

Dockerfile
```diff
@@ -38,21 +38,23 @@ COPY ./lib/ /qsim/lib/
...
 
 # Activate venv.
 ENV PATH="/test_env/bin:$PATH"
 
 # Install qsim requirements.
-# hadolint ignore=DL3042
-RUN python3 -m pip install -r /qsim/requirements.txt && \
-    python3 -m pip install -r /qsim/dev-requirements.txt
+# hadolint ignore=DL3013
+RUN python3 -m pip install --no-cache-dir --upgrade pip && \
```
pavoljuhas: I think upgrade of pip should be unnecessary after `venv --upgrade-deps` above.

Dockerfile
```diff
@@ -38,21 +38,23 @@ COPY ./lib/ /qsim/lib/
 COPY ./pybind_interface/ /qsim/lib/
 COPY ./qsimcirq_tests/ /qsim/qsimcirq_tests/
 COPY ./requirements.txt /qsim/requirements.txt
-COPY ./dev-requirements.txt /qsim/dev-requirements.txt
+COPY ./pyproject.toml /qsim/pyproject.toml
 
 # Create venv to avoid collision between system packages and what we install.
-RUN python3 -m venv --upgrade-deps test_env
+RUN python3 -m venv --upgrade-deps test_env && \
+    . test_env/bin/activate
```
pavoljuhas: I am not very familiar with docker, but if every RUN invokes a new shell, the activate has no effect and can be dropped:


```suggestion
```

docs/install_qsimcirq.md
```diff
@@ -10,18 +10,24 @@ directly in C++ code without building and installing the qsimcirq interface.
...
-[`dev-requirements.txt`](https://github.com/quantumlib/qsim/blob/main/dev-requirements.txt)
-file. You can install them with `pip3 install -r dev-requirements.txt` or
-`pip3 install qsimcirq[dev]`.
+[`pyproject.toml`](https://github.com/quantumlib/qsim/blob/main/pyproject.toml)
+file. Using pip version 25.1 or higher, you can install them with the following
+commands:
+
+```shell
+pip install -r requirements.txt
+pip install --group dev
```
pavoljuhas, line 29: consider joining these in one pip command.

pyproject.toml
```diff
@@ -12,28 +12,135 @@
...
 dependency-versions = "latest"
 enable = ["cpython-prerelease"]
 environment.PIP_PREFER_BINARY = "1"
 # Due to package & module name conflict, temporarily move it away to run tests:
-before-test = "mv {package}/qsimcirq /tmp"
-test-command = "pytest -s -v {package}/qsimcirq_tests/qsimcirq_test.py && mv /tmp/qsimcirq {package}"
+before-test = "pip install --group dev && mv {package}/qsimcirq /tmp"
+test-command = """
+pytest -s -v {package}/qsimcirq_tests/qsimcirq_test.py &&
+mv /tmp/qsimcirq {package}
```
pavoljuhas, line 134: If this is to prevent the import of `{package}/qsimcirq`, it should be doable without moving it away and back -

```suggestion
cd {package}/qsimcirq_tests && pytest -s -v qsimcirq_test.py
```

(should probably be in separate PR if you'd like to add it in).

requirements.txt
```diff
@@ -1,3 +1,17 @@
...
 absl-py
 cirq-core~=1.0
-numpy>=1.26.0
+numpy>=1.26.0,<2.0; python_version < '3.11'
+numpy>=2.0; python_version >= '3.11'
+
+# These are needed because installing qsimcirq in some environments may require
+# pip to compile Pybind for that specific platform:
+cmake~=3.28.1
+pybind11[global]
```
pavoljuhas: Should these be in `[build-system].requires` instead?

`cmake` is quite a heavy weight of about 73Mb, but is unlikely needed at qsimcirq runtime.

.github/workflows/ci.yaml
mhucka: I agree, and I had tried that IIRC got an error from pip. I tried it again, and there is no error, so ¯\_(ツ)_/¯. I'll change it to one command.

pyproject.toml
mhucka: That's a great point. That code moving it back and forth predates my joining QAI – I just left it as-is. I'll change it in a separate PR because I think there are other places where the same thing is done.

requirements.txt
mhucka: I recall having some trouble in some combination of environments and build commands when they weren't in requirements.txt. However, I worked on it some more, and I think it can be solved by putting the cmake & pybind into both the `[build-system].requires` and the `dev` group. The next push will have that.

Dockerfile
mhucka: Quite right.

Dockerfile
mhucka: Good catch. The effect of activation is done a few lines below in the `ENV PATH...` setting.

pavoljuhas (approved):

PR #992, opened 2026-01-06
Add `-mbmi2` flag when `-mavx2` is used in Makefiles

mhucka:
It looks like it's possible to use the `-mbmi2` flag when `-mavx2` is used, based on documentation and on testing locally. This was done in some Makefiles but not consistently. This PR adds the flag to Makefiles where it was missing, plus also in one BUILD file.

### DIFF
--- a/pybind_interface/avx2/CMakeLists.txt
+++ b/pybind_interface/avx2/CMakeLists.txt
@@ -20,15 +20,15 @@ if(WIN32)
     # Add /O2 to any configuration that is NOT Debug.
     # This prevents a conflict with /RTC1 in DEBUG builds.
     add_compile_options($<$<NOT:$<CONFIG:Debug>>:/O2>)
-else()
-    include(CheckCCompilerFlag)
-    check_c_compiler_flag("-mbmi2" COMPILER_HAS_BMI2)
-    if(COMPILER_HAS_BMI2)
-        add_compile_options("-mbmi2")
-        add_compile_definitions(HAVE_BMI2)
-    endif()
-
+elseif(LINUX)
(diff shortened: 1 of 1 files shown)


pybind_interface/avx2/CMakeLists.txt
```diff
@@ -21,6 +21,13 @@ if(WIN32)
     # This prevents a conflict with /RTC1 in DEBUG builds.
     add_compile_options($<$<NOT:$<CONFIG:Debug>>:/O2>)
 else()
+    include(CheckCCompilerFlag)
+    check_c_compiler_flag("-mbmi2" COMPILER_HAS_BMI2)
```
sergeisakov: That just checks if the compiler accepts this flag. BMI2 might still not be supported by CPU. I think this is okay for now. Perhaps we should have two AVX2 versions (with BMI2 and without) and detect one at runtime as we now detect SSE/AVX2/AVX512. We can add the second version later.

mhucka: > That just checks if the compiler accepts this flag. BMI2 might still not be supported by CPU. I think this is okay for now. Perhaps we should have two AVX2 versions (with BMI2 and without) and detect one at runtime as we now detect SSE/AVX2/AVX512. We can add the second version later.

Yeah, I confess I used the compiler flags because I couldn't find a cmake construct to do it properly. Of course you're right, it's incorrect and lazy.

So I searched for a better way. I found some alternatives, but they are more involved and IMHO better left to a separate PR. The simplest solution for this PR seems to be to use grep as is done in tests/Makefile. The latest push to this PR has the change to the CMakeLists.txt to do it that way instead of checking compiler flags.

@sergeisakov sorry to ask you to take yet another look …

PR #1002, opened 2026-01-20
Set LTO option only if it's supported and the build is not a debug build

mhucka:
Previously, the CMakeLists.txt files added `-flto` to some of the compilation targets unconditionally. The addition should only be done if the compiler supports LTO, and it shouldn't be done if the cmake configuration is the debug configuration.

This PR moves the detection of availability of LTO to the top-level CMakeLists.txt file, and defines a CMake macro to set it in the sub-CMakeLists.txt files. The macro avoids repeating some gnarly CMake syntax all over the place.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,15 @@ else()
     endif()
 endif()
 
+# Use OpenMP if it can be found.
+find_package(OpenMP COMPONENTS CXX NO_POLICY_SCOPE)
+add_library(qsim_openmp_config INTERFACE)
+if(OpenMP_CXX_FOUND)
+    target_link_libraries(qsim_openmp_config INTERFACE OpenMP::OpenMP_CXX)
+else()
+    message(STATUS "${MSG_PREFIX} cannot use thread parallelization")
+endif()
+
 # Add optimization flags to any configuration that is NOT Debug. Note: CMake
 # will automatically use the appropriate debug flags (e.g., /Od and /Zi for
 # MSVC, or -O0 -g for GCC/Clang), so there's no need to do that part.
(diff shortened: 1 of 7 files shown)


pybind_interface/avx2/CMakeLists.txt
```diff
@@ -50,3 +50,9 @@ include(../GetPybind11.cmake)
 pybind11_add_module(qsim_avx2 pybind_main_avx2.cpp)
 
 target_link_libraries(qsim_avx2 PUBLIC OpenMP::OpenMP_CXX)
+
+if(HAVE_IPO)
+    set_property(TARGET qsim_avx2 PROPERTY
+        INTERPROCEDURAL_OPTIMIZATION $<NOT:$<CONFIG:Debug>>
+    )
+endif()
```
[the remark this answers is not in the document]
mhucka: Done.

pybind_interface/avx2/CMakeLists.txt
```diff
@@ -17,8 +17,11 @@ project(qsim)
 
 if(MSVC)
     add_compile_options(/arch:AVX2 /openmp)
-else()
-    add_compile_options(-mavx2 -mfma -flto=auto)
+    # Add /O2 to any configuration that is NOT Debug.
+    # This prevents a conflict with /RTC1 in DEBUG builds.
+    add_compile_options($<$<NOT:$<CONFIG:Debug>>:/O2>)
+elseif(LINUX)
```
sergeisakov: Should it be `else()` as Gemini suggested in https://github.com/quantumlib/qsim/pull/1000.

mhucka: Yes, you're right. Thanks for catching that.

PR #1004, opened 2026-01-21
Add a `.geminiignore` file and improve the Gemini style guide

mhucka:
This adds a `.geminiignore` file to make Gemini CLI ignore a variety of files that are not useful for it to look into.

This PR also adds more information to the `.gemini/styleguide.md` file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,15 @@ else()
     endif()
 endif()
 
+# Use OpenMP if it can be found.
+find_package(OpenMP COMPONENTS CXX NO_POLICY_SCOPE)
+add_library(qsim_openmp_config INTERFACE)
+if(OpenMP_CXX_FOUND)
+    target_link_libraries(qsim_openmp_config INTERFACE OpenMP::OpenMP_CXX)
+else()
+    message(STATUS "${MSG_PREFIX} cannot use thread parallelization")
+endif()
+
 # Add optimization flags to any configuration that is NOT Debug. Note: CMake
 # will automatically use the appropriate debug flags (e.g., /Od and /Zi for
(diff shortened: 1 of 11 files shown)


.geminiignore
```diff
@@ -0,0 +1,54 @@
...
+/tests/googletest/
+
+# Downloaded by our build files. Not useful for Gemini to work on.
+/eigen/
+
+# Python-specific items.
+*.coverage
+*.coverage.*
+*.egg-info/
+*.py,cover
```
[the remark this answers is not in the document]
mhucka: This is incorrect. Files `*.py.cover` are generated when you asked for annotated files.

.geminiignore
```diff
@@ -0,0 +1,54 @@
...
+*.pyc
+*__pycache__*
```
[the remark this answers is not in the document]
mhucka: Good point. Fixed.

.geminiignore
```diff
@@ -0,0 +1,54 @@
...
+.ruff_cache/
+
+# Jupyter-specific items.
+*/.ipynb_checkpoints/
+ipython_config.py
+
+# Miscellaneous cache files.
+.cache/
+
+# Miscellaneous common editor backup files and temp files.
```
sergeisakov: Those files should not get into the repo in general, right?

mhucka: That's correct. Ah, you're probably pointing out that the same exclusions should be made in .gitignore. Yes, I think you're right. I'll make the change.

mhucka: OK, the latest moves most things into `.gitignore`, and adds a couple more items that came to mind while doing that. It's a better arrangement – thanks for raising the question.

PR #1006, opened 2026-01-27
Fix scan alert #151 by testing user-provided CMake arguments in `setup.py`

mhucka:
[Code scanning alert #151]( https://github.com/quantumlib/qsim/security/code-scanning/151) reports that the `subprocess.run()` invocation includes values (from `cmake_args`) that are user-provided and thus insecure.

> _Code that passes user input directly to exec, eval, or some other library routine that executes a command, allows the user to execute malicious code._
>
>```python
>        additional_cmake_args = os.environ.get("CMAKE_ARGS", "")
>        if additional_cmake_args:
>            cmake_args += additional_cmake_args.split()
>```

The only way that user-provided values could be snuck in here is via the environment variable `CMAKE_ARGS`. The types of CMake arguments that make sense in this context are flags, and all of those begin with a dash. As a (admittedly limited) form of protection against this vulnerability, this PR adds a test of the values from CMAKE_ARGS to verify they all begin with a dash character.

This does restrict the form of the flags that can be passed in `CMAKE_ARGS`. Technically, CMake understands two forms: `-DCMAKE_CXX_STANDARD=17` and `-D CMAKE_CXX_STANDARD=17`. The change in this PR means the flags always have to be of the first form, which seems to be a small inconvenience but not something that prevents any flags from being written.

Included in this PR are some new unit tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,10 +86,10 @@ class CMakeBuild(build_ext):
         # This is e.g. used by cibuildwheel to force a certain C++ standard.
         additional_cmake_args = os.environ.get("CMAKE_ARGS", "")
         for arg in additional_cmake_args.split():
-            if not arg.startswith("-"):
+            if not arg.startswith("-D"):
                 raise RuntimeError(
                     f"The value '{arg}' in the environment variable CMAKE_ARGS "
-                    "is invalid; all arguments must begin with a dash (-)."
+                    "is invalid; only definition arguments starting with '-D' are allowed."
                 )
             cmake_args.append(arg)
 


qsimcirq_tests/setuppy_test.py
```diff
@@ -0,0 +1,78 @@
...
+        return subprocess.run(
+            [sys.executable, "setup.py", "build_ext"],
+            env=env,
+            capture_output=True,
+            text=True,
+            cwd=str(root_dir),
+        )
+    except FileNotFoundError as e:
+        # This Dummy class emulates what subprocess.run() returns.
+        class Dummy:
```
pavoljuhas, line 37: As a general comment, term dummy is problematic and should be avoided in naming things, go/respectful-words.

mhucka: Oh yes, quite right. Thank you for pointing that out.

PR #1015, opened 2026-02-05
Document NumPy 1 interaction with max simulation size

### LINKED ISSUE #863: Document NumPy 1 interaction with max simulation size
### What is the task?

A user reported being unable to simulate models with more than 32 qubits. It turned out they were using NumPy 1.26, and updating to version 2.x made the problem go away. This is something we should document somewhere.

### (Optional) Do you have ideas or preferences for the approach?

_No response_

### How urgent is this for you?

P1 – needed by the next release

kris524:
Fixes #863.


docs/choose_hw.md
```diff
@@ -311,6 +311,13 @@ depth beyond 20 qubits.
     *   The total small circuits runtime overhead for an N qubit circuit
         depends on the circuit depth and on N. The overhead can be large enough to
         conceal the  $2^N$ growth in runtime.
+*   NumPy version matters for large simulations:
```
mhucka:
```suggestion
*   The NumPy version matters for large simulations:
```

docs/choose_hw.md
```diff
@@ -311,6 +311,13 @@ depth beyond 20 qubits.
...
+    *   Some users have reported being unable to simulate circuits with more than
+        ~32 qubits when using NumPy 1.26.
+    *   Updating NumPy to version 2.x resolved the issue in these cases.
```
mhucka:
```suggestion
    *    NumPy 1.x defines a constant `NPY_MAXDIMS`, which is set to 32 in the NumPy source code. Many internal NumPy structures, such as those for iteration and coordinate tracking, use fixed-size buffers based on this value. Although the C++ core of qsim is not affected by this (since it does not use NumPy), this NumPy limit prevents state vectors from having more than 32 qubits when using Python. Attempting to use 33 or more qubits results in an error.
    *    NumPy 2.x increased this limit to 64 dimensions, and qsim can work with more than 32 qubit when NumPy 2.x is used.
```

mhucka (changes requested):
Thank you for this contribution. Please see the suggested edits in the review. If you are satisfied with the suggestions, you can use GitHub's feature for accepting edits directly, to make it easier to amend the PR.

docs/choose_hw.md
```diff
@@ -311,6 +311,13 @@ depth beyond 20 qubits.
...
+    *   If you encounter unexpected failures, memory errors, or hard limits on
+        the number of qubits, ensure you are running a recent NumPy version
+        before further debugging.
```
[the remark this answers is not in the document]
mhucka: This is fixed in the most recent edits.

mhucka (approved):

PR #1034, opened 2026-04-03
Fix multiple issues in size-labeler.sh

mhucka:
Changes:

* The `IGNORED` array in `size-labeler.sh` contained glob patterns that were never used in this repository. This PR removes that and the associated code.

* Previously, only a max of 100 files were fetched from GitHub, which might have been sufficient for all cases. This PR adds support for using a paging parameter to get more than 100 files back from the GitHub API.

* The code analyzing the changes was inefficient. This rewrites the code to be more efficient and shorter. The solution was developed by Pavol Juhas and uses `gh`.

* The label `Size:XS` had inconsistent capitalization compared to the other labels. Fixed.


dev_tools/ci/size-labeler.sh
```diff
@@ -114,7 +115,7 @@ function compute_changes() {
     local response
     local change_info
     local -r keys_filter='with_entries(select([.key] | inside(["changes", "filename"])))'
-    response="$(api_call "pulls/${pr}/files")"
+    response="$(api_call "pulls/${pr}/files?per_page=100")"
```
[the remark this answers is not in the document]
mhucka: Fixed in the latest push.

dev_tools/ci/size-labeler.sh
pavoljuhas: None of these was ever present in the repo.  I suggest to remove the IGNORED array and the related skip-file logic.

Alternatively, we can replace it with and extended [regular expression for jq](https://jqlang.org/manual/#regular-expressions) see below why.

dev_tools/ci/size-labeler.sh
```diff
@@ -110,30 +110,31 @@ function api_call() {
...
     local -r pr="$1"
+    local page=1
+    local total_changes=0
+    while true; do
+        local response
+        response="$(api_call "pulls/${pr}/files?per_page=100&page=${page}")"
+
+        if [[ "$(jq_stdin '. | length' <<<"${response}")" -eq 0 ]]; then
+            break
+        fi
```
pavoljuhas, line 113: I suggest to replace all the pagination logic with `gh api --paginate` and let it do the next-page calls for us.  We can also let the `gh` sum up change counts in a jq expression, for example,

```shell
gh api --paginate /repos/quantumlib/qsim/pulls/965/files \
   --jq="[0, .[].changes] | add"
```
(the leading 0 in the array is for getting an empty array to sum to 0 rather than null)

Extending the expression with some filter for an IGNORED regex pattern would be straightforward, but we should leave it for later when (if ever) it becomes necessary.

pavoljuhas (changes requested):
If we are getting into these weeds, we can get the same using `gh api --paginate` instead.

I would also remove the IGNORED array and logic as those patterns do not match anything in the qsim repo.

dev_tools/ci/size-labeler.sh
mhucka: Because I was trying to save time and effort by using the same size-labeler.sh script everywhere.

dev_tools/ci/size-labeler.sh
mhucka: That's a clever and nicely simple solution, but note that the current code does not use `gh`. This change would introduce a dependency on `gh`. If we do that, it would be better to rewrite the script to remove the `api_call()` function and replace it with direct uses of `gh` everywhere. But I'm not sure it's a good idea to depend on `gh` for this script.

In the latest push, I simplified the `compute_changes()` function by taking advantage of the `add` trick above, and also removed `IGNORED`. Let me know what  you think.

pavoljuhas: `gh` is included in the [ubuntu-slim](https://github.com/actions/runner-images/blob/373741ffc0b04c5d6e16a2dd7bea589945f0253a/images/ubuntu-slim/ubuntu-slim-Readme.md?plain=1#L36) runner so there is nothing extra to do to use it on GHA.

AFAICT the pagination should work as is.  That said we should eventually transition to the `gh` here and in Cirq to save ourselves trouble with low-level API calls.

pavoljuhas (approved):
Please update the PR description before merge.

PR #1046, opened 2026-04-12
Micro-optimization: replace list lookup with set membership

mhucka:
In `qsimcirq/qsim_circuit.py`, changing the list `GATE_PARAMS` to a set will result in O(1) lookups. Not likely to make a big difference for the small set size, but perhaps it might shave some time off when many simulations are executed.


sergeisakov (approved):
LGTM

PR #1050, opened 2026-04-20
Add NEON simulator and statespace headers

xxie24:
This PR adds initial ARM NEON integration by introducing:

- `lib/simulator_neon.h`
- `lib/statespace_neon.h`
- `lib/simmux.h` dispatch updates for `__ARM_NEON__`

Notes:
- The NEON headers are ARM-NEON-only and use `__ARM_NEON__`.
- Naming and public structure were aligned with the SSE counterparts.


lib/statespace_neon.h
```diff
@@ -0,0 +1,441 @@
...
+      uint64_t size = MinSize(state.num_qubits()) / 8;
+      const fp_type* p = state.get();
+
+      for (uint64_t k = 0; k < size; ++k) {
+        for (unsigned j = 0; j < 4; ++j) {
+          double re = p[8 * k + j];
+          double im = p[8 * k + 4 + j];
+          norm += re * re + im * im;
+        }
+      }
```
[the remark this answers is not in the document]
sergeisakov: The reason for this specific loop was to match the loop below. `double norm = this->Norm(state);` will give a slightly different norm due to round-off errors.

lib/simulator_neon.h
```diff
@@ -0,0 +1,896 @@
...
+        } else {
+          return ExpectationValueL<4, 2>(qs, matrix, state);
+        }
+      default:
+        break;
+    }
+    return 0;
+  }
+
+  static unsigned SIMDRegisterSize() { return 4; }
```
sergeisakov: Could you format this function as
```
  /**
   * @return The size of SIMD register if applicable.
   */
  static unsigned SIMDRegisterSize() {
    return 4;
  }
```

lib/simulator_neon.h
```diff
@@ -0,0 +1,896 @@
...
+    FillIndices<H, L>(state.num_qubits(), qs, ms, xss);
+    FillMatrix<H, L, 2>(m.qmaskl, matrix, w);
+
+    const unsigned k = 2 + H;
+    const unsigned n = state.num_qubits() > k ? state.num_qubits() - k : 0;
+    const uint64_t size = uint64_t{1} << n;
+
+    using Op = std::plus<std::complex<double>>;
+    return for_.RunReduce(size, f, Op(), w, ms, xss, qs[0], state.get());
+  }
```
sergeisakov, line 895: Could you add an empty line between `}` and `For for_;`?

sergeisakov (changes requested):
Thank you for adding this. In general, it looks good to me. I have just two minor requests.

sergeisakov (approved):
LGTM

PR #1055, opened 2026-05-07
Have Bazel disable GCC SFrame option if possible

mhucka:
This adds the flag `-Wa,--gsframe=no` to the Bazel configuration for Linux systems, to deal with an error that happens with GCC 15.2. A similar addition was made to the Makefiles in a past PR.


sergeisakov (approved):
LGTM

PR #1061, opened 2026-05-10
Add a test for an uncovered case involving `add_op_to_opstring`

mhucka:
Added a new test case `test_add_op_to_opstring_controlled_gate_error` to `qsimcirq_tests/qsimcirq_test.py` to verify that `add_op_to_opstring` correctly raises a `ValueError` when a controlled operation is passed.

### DIFF
--- a/qsimcirq_tests/qsimcirq_test.py
+++ b/qsimcirq_tests/qsimcirq_test.py
@@ -858,7 +858,7 @@ def test_decompose_to_matrix_gates():
 
 def test_add_op_to_opstring_controlled_gate_error():
     q0, q1 = cirq.LineQubit.range(2)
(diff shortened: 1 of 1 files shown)


sergeisakov (approved):
LGTM. However, it seems the PR description is somewhat misleading. It seems Gemini's comment is also misleading. Both `cirq.X(q0).controlled_by(q1)` and `cirq.CX(q1, q0)` return `CNOT`, which is technically a controlled gate but its type is `GateOperation`, not `ControlledOperation`, and in both cases the kind would be `kCX`. So it triggers the gate kind validation error. On the other hand, for example, `cirq.Y(q0).controlled_by(q1)` returns a controlled operation and triggers the qubit count validation error.

PR #1075, opened 2026-05-26
Add cache-friendly qubit index remapping

xxie24:
The remap scores circuit qubits by gate usage, places frequently used qubits on lower physical state-index bits, and rewrites gates so simulation preserves the original logical circuit semantics. This improves memory locality for hot qubits during gate application.

Summary:

* Adds -r / cache_local_remap support.
* Builds a logical-to-physical qubit map from gate usage.
* Remaps gate targets, controls, and matrix ordering.
* Supports fused-gate execution by scoring fused gates and rebuilding fused metadata after remap.
* Remaps requested output amplitudes back to logical qubit order.

Benchmark note:
 * Measured about 9% runtime improvement on the q30 circuit in local testing.

### DIFF
--- a/apps/qsim_base.cc
+++ b/apps/qsim_base.cc
@@ -24,13 +24,14 @@
 #include "../lib/fuser_mqubit.h"
 #include "../lib/io_file.h"
 #include "../lib/operation.h"
+#include "../lib/qubit_remap.h"
 #include "../lib/run_qsim.h"
 #include "../lib/simmux.h"
 #include "../lib/util_cpu.h"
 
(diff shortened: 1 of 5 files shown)


sergeisakov (commented):
Thank you for the PR! I took a look at the performance impact, but I'm not quite seeing the 9% runtime improvement on my end. Depending on the circuit and platform, my benchmarks show results fluctuating within about ±5% (runtime can increase). Because the gains seem marginal here, it might be worth exploring alternative strategies for optimizing performance.

xxie24:
close this, the main work is shifted into the https://github.com/quantumlib/qsim/pull/1092 which has more effective optimization.

PR #1076, opened 2026-05-28
DO NOT MERGE - viability check of cirq-8054

pavoljuhas:
This PR is intended to store the memory usage data from valgrind
to check if https://github.com/quantumlib/Cirq/pull/8054 saves
memory in qsimcirq simulations.

- Use new flag added in quantumlib/Cirq#8054
- Example qsim simulation to monitor memory use
- Memory usage results before and with cirq-8054

DO NOT MERGE


pavoljuhas:
NB: The lint check failure [here](https://github.com/quantumlib/qsim/actions/runs/26589975157/job/78345885254?pr=1076) is expected, as this PR uses a new flag from https://github.com/quantumlib/Cirq/pull/8054.

PR #1090, opened 2026-07-06
fix: support linux/arm64 builds

Cynocracy:
The previous directives supported macos arm but did not correctly trigger off of macosx, this applies the same codepaths for linux arm, supporting them natively.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,13 +34,6 @@ else()
     message(STATUS "${MSG_PREFIX} did not detect Apple Silicon")
 endif()
 
-# SSE/AVX pybind modules use x86 intrinsics; only build them on x86.
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(x86_64|AMD64|i386|i486|i586|i686)$")
-    set(QSIM_X86 TRUE)
-else()
-    set(QSIM_X86 FALSE)
-endif()
-
 if(CMAKE_CUDA_COMPILER)
     enable_language(CUDA)
     message(STATUS "${MSG_PREFIX} found CUDA compiler "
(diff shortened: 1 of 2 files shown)


Cynocracy:
I /think/ the failures are unrelated to the changes I've made here, but let me know if they should be looked into :saluting_face: 

For context this change is useful for me to be able to support multi-arch docker images for qsim across arm/amd, the support for macos seems to prove out that the CPU is OK, this resulted from me looking into what would be needed to close the gap. No urgency on my part so if there are requested changes/improvements I'm all ears!

Cynocracy:
@mhucka does this change make sense to you? It would help my container using mac colleagues greatly so I'd love to see it included, happy to iterate however makes sense :)

mhucka:
@Cynocracy Thank you for this work, and sorry for the late reply. Yes, that failure on Windows is something new that's showing up in other CI failures. Something in the GitHub Windows environment (or maybe a third-party dependency brought in by qsim) must have changed. I need to dig in and fix the root cause.

Cynocracy:
No worries at all! Thanks for letting me know :) If I can be of assistance, happy to take a stab at debugging.

CMakeLists.txt
```diff
@@ -100,7 +107,9 @@ if(NOT CMAKE_APPLE_SILICON_PROCESSOR)
     elseif(has_hipcc)
         add_subdirectory(pybind_interface/hip)
     endif()
+endif()
 
+if(QSIM_X86)
```
mhucka: It looks like `QSIM_X86` is only used here in this one test. I think for now, we should simply things and skip the creation of a new variable, and simply in-line the test here, and remove the new lines 37-43 above (but move the comment here):
```suggestion
if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(x86_64|AMD64|i386|i486|i586|i686)$")
```

mhucka (changes requested):
A trivial change suggestion; otherwise, looking good.

mhucka:
> No worries at all! Thanks for letting me know :) If I can be of assistance, happy to take a stab at debugging.

The CI failure on Windows should be fixed now with a PR merged today. When you make changes and update the branch on this PR, it should pass.

CMakeLists.txt
Cynocracy: thanks! Makes sense to me :)

mhucka (approved):
Thank you for your work on this!

Cynocracy:
Thank you for the reviews and feedback! :) Excited to try this change out

mhucka:
If you have access to AMD hardware and have additional improvements for qsim, we'd be happy to take them.

PR #1092, opened 2026-07-23
Add gate-batched cache-local runner across SIMD backends

xxie24:
[build_gate_batch.sh](https://github.com/user-attachments/files/30323927/build_gate_batch.sh)

## Summary

This PR adds an experimental gate-batched, cache-local state-vector runner for the NEON, SSE4.2, AVX2, and AVX-512 backends.

The runner groups causally executable gates, remaps their logical qubits into a cache-sized physical-qubit block, fuses the selected gates once, and applies them to every state block. It also includes adaptive block sizing for small circuits, optional topology-aware SMT teams, and Linux NUMA interleaving for large state-vector allocations.

## Algorithm

The state vector is divided into contiguous blocks of `2^L` amplitudes. Each block represents the state space of `L` physical qubits, while the remaining `n-L` qubits have a fixed bit assignment.

While circuit gates remain:

1. Select a causally executable gate batch whose logical qubits fit in the `L` block-local physical positions.
2. Remap the required logical qubits into those positions. All disjoint qubit-position swaps are applied together in one state-vector pass.
3. Fuse the selected gates once to produce the executable gates for the batch.
4. Apply every executable gate to every state block while that block is cache-resident.

The default execution path follows the proposal's three-level loop:

```c++
while (gates_remain) {
  const auto plan = PlanNextGateBatch();
  ApplySwapsToState(plan);
  const auto executable_gates = FuseBatchGates(plan);

  #pragma omp parallel for schedule(dynamic, 1)
  for (int64_t block_index = 0;
       block_index < num_state_blocks;
       ++block_index) {
    auto block = GetStateBlock(block_index);

    for (const auto& gate : executable_gates) {
      sequential_simulator.ApplyGate(gate, block);
    }
  }
}
```

One worker normally handles one contiguous block and applies all executable gates using a sequential SIMD simulator. The simulator's existing architecture-specific kernel retains the innermost local-index loop.

On Linux, optional team-SMT execution assigns the two logical CPUs belonging to one physical core to the same block. Each sibling processes a disjoint portion of the simulator kernel and the pair synchronizes between gates. CPU topology is discovered at runtime from the process affinity and sysfs; workers are pinned explicitly rather than assuming adjacent OpenMP thread IDs are siblings.

`L` is treated as an upper bound and may be reduced for small circuits to expose enough blocks for parallel execution. Frequently used logical qubits are initially placed in a fixed block-local region. SIMD lane positions remain pinned because remapping moves whole SIMD chunks rather than individual lanes.

`QubitMappedState` retains the final logical-to-physical mapping so callers can avoid a final identity-restoration pass.

On multi-node Linux systems, large state-vector allocations use `MPOL_INTERLEAVE` when multiple compute nodes are available. Unsupported platforms, single-node systems, and failed NUMA-policy calls retain the normal aligned-allocation behavior.

## Performance

Times are in seconds and lower is better. Speedup is base-runner time divided by gate-batch-runner time. Results should only be compared within a row/platform.

### Full `circuit_q30` measurements — August 19, 2026

These measurements use the full 1617-raw-gate circuit and compare the best configuration measured for each runner. They predate the team-SMT and NUMA-interleaving commits, so they should not be treated as measurements of the final PR head.

| Platform | SIMD backend | Base best configuration | Base (s) | Gate-batch best configuration | Gate-batch total (s) | Gate application (s) | Remap (s) | Speedup |
|---|---|---|---:|---|---:|---:|---:|---:|
| Google C4 / Intel | AVX-512 | `f=4`, SMT on (192 threads) | **7.96** | `f=4, L=18`, SMT off (96 threads) | **8.18** | 4.12 | 4.05 | **0.97x** |
| Google C4 / Intel | SSE4.2 | `f=4`, SMT on (192 threads) | **10.67** | `f=3, L=18`, ordinary SMT (192 threads) | **11.97** | 7.52 | 4.45 | **0.89x** |
| Google C4D / AMD Turin | AVX-512 | `f=4`, SMT on (96 threads) | **10.53** | `f=3, L=17`, SMT off (48 threads) | **6.13** | 3.40 | 2.73 | **1.72x** |
| Google C4D / AMD Turin | SSE4.2 | `f=3`, SMT on (96 threads) | **17.39** | `f=3, L=17`, SMT off (48 threads) | **16.14** | 12.67 | 3.46 | **1.08x** |
| Google C4A / Axion | NEON | `f=4`, 72 physical cores | **11.00** | `f=3, L=18`, 72 physical cores | **9.99** | 6.82 | 3.17 | **1.10x** |

The results are architecture- and configuration-dependent. The largest measured gain is on AMD Turin with AVX-512. Axion shows a modest gain, while the best large Intel C4 base configuration is slightly faster than the best gate-batch configuration. Remapping remains a material part of total runtime, so improved gate locality does not automatically translate into an end-to-end speedup.

### Intel C4 team-SMT experiment — August 24, 2026

This focused experiment used an 8-physical-core/16-vCPU `c4-standard-16` Intel Xeon Platinum 8581C VM. Each core has a private 2 MiB L2, so the gate-batch runs used `L=18`. The circuit was limited with `d=30` (502 raw gates), and each result is the average of two runs with `f=3`.

| Configuration | CPU allocation | Total (s) | Gate application (s) | Remap (s) | Speedup vs 8-core base |
|---|---|---:|---:|---:|---:|
| Base runner | 8 physical cores | 17.980 | — | — | 1.00x |
| Gate-batch, physical only | 8 independent workers | 9.297 | 6.319 | 2.976 | 1.93x |
| Gate-batch, ordinary SMT | 16 independent workers | 9.257 | 6.762 | 2.493 | 1.94x |
| Gate-batch, team SMT | 8 two-sibling teams | **8.523** | **5.930** | 2.592 | **2.11x** |

Runtime topology discovery produced the sibling pairs `(0,8)`, `(1,9)`, ..., `(7,15)`. Team SMT was about 9% faster end-to-end than the 8-worker gate-batch run. Ordinary SMT made gate application slower because two siblings processed separate L2-sized blocks; assigning both siblings to one shared block recovered that loss and improved the gate phase.

This short `d=30` experiment is not directly comparable to the full-circuit table above. A full latest-head sweep, including NUMA interleaving and team SMT on the large C4/C4D machines, remains future work.

## Validation

- Compared logical amplitudes/full state vectors with the existing qsim runner, accounting for the retained qubit mapping.
- Tested multiple block sizes and fusion sizes.
- Tested NEON, SSE4.2, AVX2, and AVX-512 builds.
- Verified generated NEON and AVX-512 SIMD instructions.
- Verified identical q24 amplitudes between physical-only and team-SMT execution on Intel C4.
- Verified runtime Intel sibling discovery and explicit worker pinning.
- Verified that unavailable/restricted SMT siblings are rejected rather than paired incorrectly.
- Tested the NUMA allocation change with `bazel test //tests:all` on Linux.
- Observed numerical differences around `1e-9` in regression testing.

## Remaining work

- Re-run the full q30 performance matrix at the latest PR head.
- Base adaptive block sizing on the number of block teams rather than total logical threads.
- Separate physical-core remap workers from the larger gate-thread pool.
- Add pause/backoff behavior to the team synchronization barrier.
- Add mocked topology tests, OpenMP CI coverage, and topology discovery beyond Linux.

### DIFF
--- a/apps/BUILD
+++ b/apps/BUILD
@@ -33,6 +33,16 @@ cc_binary(
 )
 
 cc_binary(
+    name = "qsim_gate_batch",
+    srcs = ["qsim_gate_batch.cc"],
+    copts = gsframe_copts,
+    deps = [
+        "//lib:run_qsim_gate_batch",
+        "//lib:run_qsim_lib",
+    ],
+)
+
+cc_binary(
     name = "qsim_von_neumann",
     srcs = ["qsim_von_neumann.cc"],
     copts = gsframe_copts,
--- a/apps/Makefile
+++ b/apps/Makefile
@@ -1,6 +1,8 @@
 CXX_TARGETS = $(shell find . -maxdepth 1 -name '*.cc')
 CXX_TARGETS := $(CXX_TARGETS:%.cc=%.x)
 
+OUTPUT_DIR ?= .
+
 CUDA_TARGETS = $(shell find . -maxdepth 1 -name '*cuda.cu')
 CUDA_TARGETS := $(CUDA_TARGETS:%cuda.cu=%cuda.x)
 
@@ -15,6 +17,12 @@ HIP_TARGETS := $(HIP_TARGETS:%cuda.cu=%hip.x)
 
 .PHONY: qsim
 qsim: $(CXX_TARGETS)
+
+.PHONY: gate-batch
+gate-batch:
+	mkdir -p "$(OUTPUT_DIR)"
+	$(CXX) -o "$(OUTPUT_DIR)/qsim_gate_batch.x" qsim_gate_batch.cc $(CXXFLAGS) $(ARCHFLAGS)
+	$(CXX) -o "$(OUTPUT_DIR)/qsim_base.x" qsim_base.cc $(CXXFLAGS) $(ARCHFLAGS)
 
 .PHONY: qsim-cuda
 qsim-cuda: $(CUDA_TARGETS)
--- a/apps/qsim_gate_batch.cc
+++ b/apps/qsim_gate_batch.cc
@@ -10,8 +10,12 @@
 
 #include <limits>
 #include <string>
+#include <utility>
+#include <vector>
 
 #include "../lib/circuit_qsim_parser.h"
+#include "../lib/cooperative_for.h"
+#include "../lib/cpu_thread_topology.h"
 #include "../lib/formux.h"
 #include "../lib/fuser_mqubit.h"
 #include "../lib/gates_qsim.h"
@@ -27,6 +31,7 @@ struct Options {
   unsigned maxtime = std::numeric_limits<unsigned>::max();
   unsigned seed = 1;
   unsigned num_threads = 1;
+  unsigned inner_threads = 1;
   unsigned max_fused_size = 3;
   unsigned block_qubits = 19;
   unsigned verbosity = 0;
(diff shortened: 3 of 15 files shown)


mhucka:
@xxie24 Thank you for this work. I know it's still a draft PR, but I wanted to mention that a recent failure in the CI workflow has now been fixed. I took the liberty of updating your branch to have the latest changes (I hope that's okay), and now the PR passes all the Ci tests.

xxie24:
> @xxie24 Thank you for this work. I know it's still a draft PR, but I wanted to mention that a recent failure in the CI workflow has now been fixed. I took the liberty of updating your branch to have the latest changes (I hope that's okay), and now the PR passes all the Ci tests.

cool. thanks.

